PR #150, opened 2023-10-27
Supports advanced indexing of `MultiNestedTensor` (part 1)

weihua916:
- Supports row slicing
- Supports a column integer indexing.
- Refactor and code clean up.

Subsequent PRs:
- Supports column advanced indexing (slicing, List indexing)
- Supports concatenation along row and column.
- Increase test coverage.

### DIFF
--- a/test/data/test_multi_nested_tensor.py
+++ b/test/data/test_multi_nested_tensor.py
@@ -21,12 +21,13 @@ def assert_equal(tensor_mat: List[List[Tensor]],
 def test_multi_nested_tensor_basic():
     num_rows = 8
     num_cols = 10
+    max_value = 100
     tensor_mat = []
     for _ in range(num_rows):
         tensor_list = []
         for _ in range(num_cols):
             length = random.randint(0, 10)
-            tensor_list.append(torch.randint(0, 100, size=(length, )))
+            tensor_list.append(torch.randint(0, max_value, size=(length, )))
         tensor_mat.append(tensor_list)
 
     multi_nested_tensor = MultiNestedTensor.from_tensor_mat(tensor_mat)
@@ -93,3 +94,9 @@ def test_multi_nested_tensor_basic():
     with pytest.raises(NotImplementedError):
         # TODO: Add proper test once implemented
         multi_nested_tensor[:, 4:8]
+    cloned_multi_nested_tensor = multi_nested_tensor.clone()
+
+    multi_nested_tensor.values[0] = max_value + 1.0
+    assert cloned_multi_nested_tensor.values[0] != max_value + 1.0
+    multi_nested_tensor.offset[0] = -1
+    assert cloned_multi_nested_tensor.values[0] != -1
--- a/test/data/test_multi_nested_tensor.py
+++ b/test/data/test_multi_nested_tensor.py
@@ -1,0 +1,50 @@
+import random
+
+import torch
+
+from torch_frame.data import MultiNestedTensor
+
+
+def test_multi_nested_tensor_basic():
+    num_rows = 8
+    num_cols = 10
+    tensor_mat = []
+    for _ in range(num_rows):
+        tensor_list = []
+        for _ in range(num_cols):
+            length = random.randint(0, 10)
+            tensor_list.append(torch.randint(0, 100, size=(length, )))
+        tensor_mat.append(tensor_list)
+
+    multi_nested_tensor = MultiNestedTensor.from_tensor_mat(tensor_mat)
+    assert str(multi_nested_tensor
+               ) == "MultiNestedTensor(num_rows=8, num_cols=10, device=cpu)"
+    assert multi_nested_tensor.num_rows == num_rows
+    assert multi_nested_tensor.num_cols == num_cols
+
+    # Test multi_nested_tensor[i, j] indexing
+    for i in range(-num_rows, num_rows):
+        for j in range(num_cols):
+            tensor = multi_nested_tensor[i, j]
+            assert isinstance(tensor, torch.Tensor)
+            assert torch.allclose(tensor_mat[i][j], tensor)
+
+    # Test multi_nested_tensor[i] indexing
+    for i in range(-num_rows, num_rows):
+        multi_nested_tensor_row = multi_nested_tensor[i]
+        assert multi_nested_tensor_row.num_rows == 1
+        assert multi_nested_tensor_row.num_cols == num_cols
+        for j in range(-num_cols, num_cols):
+            tensor = multi_nested_tensor_row[0, j]
+            assert isinstance(tensor, torch.Tensor)
+            assert torch.allclose(tensor_mat[i][j], tensor)
+
+    # Test multi_nested_tensor[List[int]] indexing
+    for index in [[4], [2, 2], [-4, 1, 7], [3, -7, 1, 0]]:
+        multi_nested_tensor_indexed = multi_nested_tensor[index]
+        assert multi_nested_tensor_indexed.num_rows == len(index)
+        assert multi_nested_tensor_indexed.num_cols == num_cols
+        for i, idx in enumerate(index):
+            for j in range(num_cols):
+                tensor = multi_nested_tensor_indexed[i, j]
+                assert torch.allclose(tensor_mat[idx][j], tensor)
--- a/test/data/test_multi_nested_tensor.py
+++ b/test/data/test_multi_nested_tensor.py
@@ -32,9 +32,17 @@ def test_multi_nested_tensor_basic():
 
     multi_nested_tensor = MultiNestedTensor.from_tensor_mat(tensor_mat)
     assert str(multi_nested_tensor
-               ) == "MultiNestedTensor(num_rows=8, num_cols=10, device=cpu)"
+               ) == "MultiNestedTensor(num_rows=8, num_cols=10, device='cpu')"
+
+    # Test sizes
     assert multi_nested_tensor.shape[0] == num_rows
+    assert multi_nested_tensor.size(0) == num_rows
     assert multi_nested_tensor.shape[1] == num_cols
+    assert multi_nested_tensor.size(1) == num_cols
+    with pytest.raises(ValueError, match="not have a fixed length"):
+        multi_nested_tensor.size(2)
+    with pytest.raises(IndexError, match="Dimension out of range"):
+        multi_nested_tensor.size(3)
 
     # Test multi_nested_tensor[i, j] indexing
     for i in range(-num_rows, num_rows):
@@ -110,3 +118,19 @@ def test_multi_nested_tensor_basic():
     assert cloned_multi_nested_tensor.values[0] != max_value + 1.0
     multi_nested_tensor.offset[0] = -1
     assert cloned_multi_nested_tensor.values[0] != -1
+
+
+def test_multi_nested_tensor_different_num_rows():
+    tensor_mat = [
+        [torch.tensor([1, 2, 3]),
+         torch.tensor([4, 5])],
+        [torch.tensor([6, 7]),
+         torch.tensor([8, 9, 10]),
+         torch.tensor([11])],
+    ]
+
+    with pytest.raises(
+            RuntimeError,
+            match="The length of each row must be the same",
+    ):
+        MultiNestedTensor.from_tensor_mat(tensor_mat)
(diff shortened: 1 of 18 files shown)


test/data/test_multi_nested_tensor.py
```diff
@@ -1,10 +1,22 @@
...
 import torch
+from torch import Tensor
 
 from torch_frame.data import MultiNestedTensor
 
 
+def assert_equal(tensor_mat: List[List[Tensor]],
+                 multi_nested_tensor: MultiNestedTensor):
+    assert len(tensor_mat) == multi_nested_tensor.num_rows
+    assert len(tensor_mat[0]) == multi_nested_tensor.num_cols
```
rusty1s, line 13: Can we also expose `shape`? I think `num_rows`, `num_cols` terminology for a tensor is quite confusing.

torch_frame/data/multi_nested_tensor.py
```diff
@@ -78,85 +78,161 @@ def from_tensor_mat(
...
         index: Any,
     ) -> Union['MultiNestedTensor', Tensor]:
+        def col_index(
+                multi_nested_tensor: 'MultiNestedTensor',
+                index: Union[int, Tensor, List, slice]) -> 'MultiNestedTensor':
+            r"""Supports all types of column-level advanced indexing of input
+            :class:`MultiNestedTensor` object."""
+            if isinstance(index, int):
+                return multi_nested_tensor.single_index_select(index, dim=1)
+            elif isinstance(index, Tensor):
```
rusty1s, line 96: There exists a bunch of non-tested code paths here.

torch_frame/data/multi_nested_tensor.py
```diff
@@ -78,85 +78,161 @@ def from_tensor_mat(
     def __repr__(self) -> str:
         name = ' '.join([
             f"{self.__class__.__name__}(num_rows={self.num_rows},",
-            f"num_cols={self.num_cols},", f"device={self.values.device})"
+            f"num_cols={self.num_cols},", f"device={self.device})"
         ])
-
         return name
 
     def __getitem__(
```
rusty1s, line 85: This method here is super complex. Can we refactor it?

torch_frame/data/multi_nested_tensor.py
```diff
@@ -78,85 +78,161 @@ def from_tensor_mat(
...
+                return multi_nested_tensor.index_select(
+                    torch.tensor(index, device=multi_nested_tensor.device),
+                    dim=1)
+            elif isinstance(index, slice):
+                # TODO: Support this
+                return multi_nested_tensor.slice(index, dim=1)
+            else:
+                raise NotImplementedError
+
         if isinstance(index, Tuple):
```
rusty1s, line 110: wouldn't it be easier if we just iterate over `index`, and then do:
```python
out = self
for i, value in enumerate(index):
    out = out.index_select(value, dim=i)

File: torch_frame/data/multi_nested_tensor.py [hunk 5]
```diff
@@ -78,85 +78,161 @@ def from_tensor_mat(
...
-    def _row_index_select(self, index: Tensor) -> Tensor:
-        index = index.clone()
-        # Support negative indices
-        neg_idx = index < 0
-        index[neg_idx] = self.num_rows + index[neg_idx]
-        if index.min() < 0 or index.max() >= self.num_rows:
-            raise IndexError("Index out of bounds!")
+            raise RuntimeError(f"Unsupported dim={dim} for index_select.")
+
+    def row_slice(self, slice: slice) -> 'MultiNestedTensor':
```
rusty1s, line 179: this is known as narrow. Let's try to follow PyTorch terminology.

rusty1s (changes requested):

torch_frame/data/multi_nested_tensor.py
weihua916: I checked `narrow` (https://pytorch.org/docs/stable/generated/torch.narrow.html) but the arguments are different from `slice`.
It needs to pass `start` and `length`. There is no `step` and `length` is different from `end`.

torch_frame/data/multi_nested_tensor.py
weihua916: Yes, will try.

torch_frame/data/multi_nested_tensor.py
weihua916: Added a temporary test on `NotImplementedError`. Will resolve it soon.

torch_frame/data/multi_nested_tensor.py
rusty1s: That's true, but you can compute end from start and length. Overall, I just think it is nicer when we inherit PyTorch functions here so that it feels like a normal tensor.

weihua916: I see. But it cannot model the step. Shall we add narrow by calling row_slice internally?

torch_frame/data/multi_nested_tensor.py
```diff
@@ -120,83 +120,93 @@ def __repr__(self) -> str:
...
-            return MultiNestedTensor(num_rows=1, num_cols=self.num_cols,
-                                     values=values, offset=offset)
-        elif isinstance(index, Tensor) and index.ndim == 1:
-            return self.index_select(index, dim=0)
-        elif isinstance(index, List):
-            return self.index_select(
-                torch.tensor(index, device=self.values.device), dim=0)
+            if isinstance(index[0], int) and isinstance(index[1], int):
+                # Returns Tensor
+                return MultiNestedTensor.get_tensor(self, index[0], index[1])
```
rusty1s, line 136: Any reason to make this static? Why not do `self.get_value(row: int, col: int)`?

torch_frame/data/multi_nested_tensor.py
```diff
@@ -120,83 +120,93 @@ def __repr__(self) -> str:
...
+            else:
+                # Returns MultiNestedTensor
+                out = self
+                for dim, idx in enumerate(index):
+                    out = MultiNestedTensor.index(out, idx, dim)
```
rusty1s, line 141: Here again. Would prefer:
```suggestion
                    out = out.select(out, idx, dim)
```

torch_frame/data/multi_nested_tensor.py
```diff
@@ -120,83 +120,93 @@ def __repr__(self) -> str:
...
-        if dim == 0:
-            return self._row_index_select(index)
+            # Returns MultiNestedTensor
+            return MultiNestedTensor.index(self, index, dim=0)
+
+    def slice(self, slice: slice, dim: int) -> 'MultiNestedTensor':
+        if dim == 0 or dim == -3:
+            return self.row_slice(slice)
+        elif dim == 1 or dim == -2:
+            return self.col_slice(slice)
```
rusty1s, line 151: Can we add an assert here in case `dim` is not well defined?

torch_frame/data/multi_nested_tensor.py
rusty1s: Sounds good.

torch_frame/data/multi_nested_tensor.py
```diff
@@ -120,83 +120,93 @@ def __repr__(self) -> str:
...
+
+    def row_slice(self, slice: slice) -> 'MultiNestedTensor':
+        r"""Slice along row (dim=0)"""
+        if slice.step is not None and slice.step > 1:
+            # If step is larger than 1, we reuse index_select along rows.
+            idx = torch.arange(self.num_rows, device=self.device)[slice]
```
rusty1s, line 157: nit: `torch.tensor(range(self.num_rows)[slice])` would be a bit more memory friendly.

torch_frame/data/multi_nested_tensor.py
```diff
@@ -210,6 +220,95 @@ def _row_index_select(self, index: Tensor) -> Tensor:
...
+                    if index < 0 or index > self.num_rows:
+                        raise IndexError("Row index out of bounds!")
+                else:
+                    if index < 0 or index >= self.num_rows:
+                        raise IndexError("Row index out of bounds!")
+            else:
+                if index < 0:
+                    index = index + self.num_cols
+                if is_slice_end:
+                    if index < 0 or index > self.num_cols:
```
rusty1s, line 285:
```suggestion
                if is_slice_end and index < 0 or index > self.num_cols:
```

torch_frame/data/multi_nested_tensor.py
```diff
@@ -210,6 +220,95 @@ def _row_index_select(self, index: Tensor) -> Tensor:
...
+                        raise IndexError("Column index out of bounds!")
+                else:
+                    if index < 0 or index >= self.num_cols:
```
rusty1s, line 288:
```suggestion
                elif index < 0 or index >= self.num_cols:
```

torch_frame/data/multi_nested_tensor.py
```diff
@@ -210,6 +220,95 @@ def _row_index_select(self, index: Tensor) -> Tensor:
...
+            torch.cumsum(diff, dim=0, out=offset[1:])
+            return MultiNestedTensor(num_rows=self.num_rows, num_cols=1,
+                                     values=values, offset=offset)
+        else:
+            raise RuntimeError(f"Unsupported dim={dim} for index_select.")
+
+    def _to_positive_index(
+        self,
+        index: Union[int, Tensor],
+        dim: int,
```
rusty1s, line 257: There is a lot of duplicated logic here depending on `dim`. Might be easier to just add a `max_entries` argument here instead of `dim`?

torch_frame/data/multi_nested_tensor.py
```diff
@@ -210,6 +220,95 @@ def _row_index_select(self, index: Tensor) -> Tensor:
...
+                    index = index + self.num_cols
+                if is_slice_end:
+                    if index < 0 or index > self.num_cols:
+                        raise IndexError("Column index out of bounds!")
+                else:
+                    if index < 0 or index >= self.num_cols:
+                        raise IndexError("Column index out of bounds!")
+        elif isinstance(index, Tensor):
+            assert not is_slice_end
+            index = index.clone()
```
rusty1s, line 292: nit: In most cases, the clone here is unnecessary:
```python
neg_index = index[neg_idx]
if neg_index.numel() > 0:
    index = index.clone()
    index[neg_idx] = max_entries + neg_index

File: torch_frame/data/multi_nested_tensor.py [hunk 14]
```diff
@@ -210,6 +220,95 @@ def _row_index_select(self, index: Tensor) -> Tensor:
...
+                    raise IndexError("Row index out of bounds!")
+            else:
+                index[neg_idx] = self.num_cols + index[neg_idx]
+                if index.min() < 0 or index.max() >= self.num_cols:
+                    raise IndexError("Column index out of bounds!")
+        return index
+
+    @property
+    def device(self) -> torch.device:
+        return self.values.device
```
rusty1s, line 307: Seems to already exist.

torch_frame/data/multi_nested_tensor.py
```diff
@@ -230,17 +329,25 @@ def _apply(self, fn: Callable[[Tensor], Tensor]) -> 'MultiNestedTensor':
...
+
+    @property
+    def ndim(self) -> int:
         return 3
 
+    def dim(self) -> int:
+        return self.ndim
+
     @property
     def device(self) -> torch.device:
```
rusty1s, line 342: Shall we also expose `dtype`?

torch_frame/data/multi_nested_tensor.py
```diff
@@ -120,83 +120,93 @@ def __repr__(self) -> str:
...
-        index = index.clone()
-        # Support negative indices
-        neg_idx = index < 0
-        index[neg_idx] = self.num_rows + index[neg_idx]
-        if index.min() < 0 or index.max() >= self.num_rows:
-            raise IndexError("Index out of bounds!")
+            start_idx: int = self._to_positive_index(slice.start or 0, dim=0)
+            end_idx: int = self._to_positive_index(slice.stop or self.num_rows,
+                                                   dim=0, is_slice_end=True)
+            if start_idx == 0 and end_idx == self.num_rows:
```
rusty1s, line 163:
```suggestion
            if start_idx <= 0 and end_idx >= self.num_rows:
```

rusty1s (approved):

torch_frame/data/multi_nested_tensor.py
weihua916: That's a good point. Fixed.

torch_frame/data/multi_nested_tensor.py
weihua916: Good point. Done.

torch_frame/data/multi_nested_tensor.py
weihua916: We already raise RuntimeError.

PR #128, opened 2023-10-20
Add Multi-categorical stype

yiweny:
Add `multicategorical` stype

PR #151, opened 2023-10-27
Fix multicategorical stype based on review comments

yiweny:
Based on the comments in #128 , the following changes are made in this pr:
1. Added an `MULTI_COUNT` stat type for `multicategorical` stype.
2. Modified the name of `multi_categorical` stype to `multicategorical` stype.
3. Modified the mappings. Now missing values will be mapped to `-1`, no value is considered as a category and unseen values are all ignored.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
-- Added `Occurrence` stat for `multicategorical` stype ([#151](https://github.com/pyg-team/pytorch-frame/pull/151))
+- Added `MULTI_COUNT` stat for `multicategorical` stype ([#151](https://github.com/pyg-team/pytorch-frame/pull/151))
 - Added `multi_categorical` stype ([#128](https://github.com/pyg-team/pytorch-frame/pull/128))
 - Added `MultiNestedTensor` ([#149](https://github.com/pyg-team/pytorch-frame/pull/149))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- Added `Occurrence` stat for `multicategorical` stype ([#151](https://github.com/pyg-team/pytorch-frame/pull/151))
 - Added `multi_categorical` stype ([#128](https://github.com/pyg-team/pytorch-frame/pull/128))
 - Added `MultiNestedTensor` ([#149](https://github.com/pyg-team/pytorch-frame/pull/149))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
-- Added `MULTI_COUNT` stat for `multicategorical` stype ([#151](https://github.com/pyg-team/pytorch-frame/pull/151))
-- Added `multi_categorical` stype ([#128](https://github.com/pyg-team/pytorch-frame/pull/128))
+- Added `multi_categorical` stype ([#128](https://github.com/pyg-team/pytorch-frame/pull/128), [#151](https://github.com/pyg-team/pytorch-frame/pull/151))
 - Added `MultiNestedTensor` ([#149](https://github.com/pyg-team/pytorch-frame/pull/149))
 
 ### Changed
--- a/test/data/test_dataset.py
+++ b/test/data/test_dataset.py
@@ -104,7 +104,7 @@ def test_multicategorical_materialization():
     data = {'multicat_col': ['A|B', 'B|C|A', '', '', 'B', 'B|A', None]}
     df = pd.DataFrame(data)
     dataset = Dataset(df, {'multicat_col': stype.multicategorical},
-                      sep={'multicat_col': '|'})
+                      col_to_sep={'multicat_col': '|'})
     dataset.materialize()
     feat = dataset.tensor_frame.feat_dict[stype.multicategorical]
     assert torch.equal(feat[0, 0], torch.tensor([1, 0]))
--- a/test/data/test_dataset.py
+++ b/test/data/test_dataset.py
@@ -101,18 +101,20 @@ def test_converter():
 
 
 def test_multicategorical_materialization():
-    data = {'a': ['A,B', 'B,C,A', '', 'B', None]}
+    data = {'a': ['A|B', 'B|C|A', '', '', 'B', 'B|A', None]}
     df = pd.DataFrame(data)
-    dataset = Dataset(df, {'a': stype.multicategorical})
+    dataset = Dataset(df, {'a': stype.multicategorical}, sep={'a': '|'})
     dataset.materialize()
     feat = dataset.tensor_frame.feat_dict[stype.multicategorical]
     assert torch.equal(feat[0, 0], torch.tensor([1, 0], device=feat.device))
-    assert torch.equal(feat[1, 0], torch.tensor([0, 2, 1], device=feat.device))
-    assert feat[2, 0].numel() == 0
-    assert torch.equal(feat[4, 0], torch.tensor([-1], device=feat.device))
-    assert StatType.COUNT in dataset.col_stats['a']
-    assert dataset.col_stats['a'][StatType.COUNT][0] == ['B', 'A', 'C']
-    assert dataset.col_stats['a'][StatType.COUNT][1] == [3, 2, 1]
+    assert torch.equal(feat[1, 0], torch.tensor([0, 3, 1], device=feat.device))
+    assert torch.equal(feat[2, 0], torch.tensor([2], device=feat.device))
+    assert torch.equal(feat[6, 0], torch.tensor([-1], device=feat.device))
+    assert StatType.OCCURRENCE in dataset.col_stats['a']
+    assert dataset.col_stats['a'][StatType.OCCURRENCE][0] == [
+        'B', 'A', '', 'C'
+    ]
+    assert dataset.col_stats['a'][StatType.OCCURRENCE][1] == [4, 3, 2, 1]
 
 
 @pytest.mark.parametrize('with_nan', [True, False])
--- a/test/data/test_dataset.py
+++ b/test/data/test_dataset.py
@@ -110,11 +110,11 @@ def test_multicategorical_materialization():
     assert torch.equal(feat[1, 0], torch.tensor([0, 3, 1], device=feat.device))
     assert torch.equal(feat[2, 0], torch.tensor([2], device=feat.device))
     assert torch.equal(feat[6, 0], torch.tensor([-1], device=feat.device))
-    assert StatType.OCCURRENCE in dataset.col_stats['a']
-    assert dataset.col_stats['a'][StatType.OCCURRENCE][0] == [
+    assert StatType.MULTI_COUNT in dataset.col_stats['a']
+    assert dataset.col_stats['a'][StatType.MULTI_COUNT][0] == [
         'B', 'A', '', 'C'
     ]
-    assert dataset.col_stats['a'][StatType.OCCURRENCE][1] == [4, 3, 2, 1]
+    assert dataset.col_stats['a'][StatType.MULTI_COUNT][1] == [4, 3, 2, 1]
 
 
 @pytest.mark.parametrize('with_nan', [True, False])
--- a/test/data/test_dataset.py
+++ b/test/data/test_dataset.py
@@ -101,17 +101,18 @@ def test_converter():
 
 
 def test_multicategorical_materialization():
-    data = {'a': ['A|B', 'B|C|A', '', '', 'B', 'B|A', None]}
+    data = {'multicat_col': ['A|B', 'B|C|A', '', '', 'B', 'B|A', None]}
     df = pd.DataFrame(data)
-    dataset = Dataset(df, {'a': stype.multicategorical}, sep={'a': '|'})
+    dataset = Dataset(df, {'multicat_col': stype.multicategorical},
+                      sep={'multicat_col': '|'})
     dataset.materialize()
     feat = dataset.tensor_frame.feat_dict[stype.multicategorical]
-    assert torch.equal(feat[0, 0], torch.tensor([1, 0], device=feat.device))
-    assert torch.equal(feat[1, 0], torch.tensor([0, 3, 1], device=feat.device))
-    assert torch.equal(feat[2, 0], torch.tensor([2], device=feat.device))
-    assert torch.equal(feat[6, 0], torch.tensor([-1], device=feat.device))
-    assert StatType.MULTI_COUNT in dataset.col_stats['a']
-    assert (dataset.col_stats['a'][StatType.MULTI_COUNT][0] == [
+    assert torch.equal(feat[0, 0], torch.tensor([1, 0]))
+    assert torch.equal(feat[1, 0], torch.tensor([0, 3, 1]))
+    assert torch.equal(feat[2, 0], torch.tensor([2]))
+    assert torch.equal(feat[6, 0], torch.tensor([-1]))
+    assert StatType.MULTI_COUNT in dataset.col_stats['multicat_col']
+    assert (dataset.col_stats['multicat_col'][StatType.MULTI_COUNT][0] == [
         'B', 'A', '', 'C'
     ])
     assert dataset.col_stats['a'][StatType.MULTI_COUNT][1] == [4, 3, 2, 1]
--- a/test/data/test_dataset.py
+++ b/test/data/test_dataset.py
@@ -89,7 +89,7 @@ def test_converter():
     dataset = FakeDataset(
         num_rows=10,
         stypes=[stype.categorical, stype.numerical,
-                stype.multi_categorical]).materialize()
+                stype.multicategorical]).materialize()
     convert_to_tensor_frame = DataFrameToTensorFrameConverter(
         col_to_stype=dataset.col_to_stype,
         col_stats=dataset.col_stats,
@@ -100,12 +100,12 @@ def test_converter():
     assert len(tf) == len(dataset)
 
 
-def test_multi_categorical_materialization():
+def test_multicategorical_materialization():
     data = {'a': ['A,B', 'B,C,A', '', 'B', None]}
     df = pd.DataFrame(data)
-    dataset = Dataset(df, {'a': stype.multi_categorical})
+    dataset = Dataset(df, {'a': stype.multicategorical})
     dataset.materialize()
-    feat = dataset.tensor_frame.feat_dict[stype.multi_categorical]
+    feat = dataset.tensor_frame.feat_dict[stype.multicategorical]
     assert torch.equal(feat[0, 0], torch.tensor([1, 0], device=feat.device))
     assert torch.equal(feat[1, 0], torch.tensor([0, 2, 1], device=feat.device))
     assert feat[2, 0].numel() == 0
--- a/test/data/test_dataset.py
+++ b/test/data/test_dataset.py
@@ -6,6 +6,7 @@ import torch
 import torch_frame
 from torch_frame import stype
 from torch_frame.data import DataFrameToTensorFrameConverter, Dataset
+from torch_frame.data.dataset import canonicalize_col_to_sep
 from torch_frame.data.stats import StatType
 from torch_frame.datasets import FakeDataset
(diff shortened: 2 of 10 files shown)


torch_frame/data/mapper.py
```diff
@@ -129,7 +130,19 @@ def forward(
         if ser.dtype != 'object':
             raise ValueError('Multi-categorical types expect string as input')
         values = []
+        original_index = ser.index
         ser = ser.apply(self._row_to_tensor)
```
rusty1s, line 134: we are not returning a tensor here, so maybe better to rename to `self._split_by_sep` or something similar.

torch_frame/data/mapper.py
```diff
@@ -129,7 +130,19 @@ def forward(
...
+        ser = ser.explode()
+        ser = pd.merge(
+            ser.rename('data'),
+            self.index.rename('index'),
+            how='left',
+            left_on='data',
+            right_index=True,
+        ).dropna()
+        ser['index'] = ser['index'].astype('int64')
+        ser = ser.groupby(level=0)['index'].apply(list)
```
rusty1s, line 144: Why do we need to group here again? We already have the flattened output, right?

torch_frame/data/stats.py
```diff
@@ -25,6 +25,9 @@ class StatType(Enum):
     # Categorical:
     COUNT = 'COUNT'
 
+    # Multicategorical:
+    OCCURRENCE = 'OCCURRENCE'
```
rusty1s, line 29: not super happy with the naming of `COUNT`/`OCCURRENCE`. Maybe better to rename this to `MULTI_COUNT`?

torch_frame/data/stats.py
```diff
@@ -39,14 +42,14 @@ def stats_for_stype(stype: torch_frame.stype) -> List['StatType']:
...
+        elif stype == torch_frame.multicategorical:
             return [
-                StatType.COUNT,
+                StatType.OCCURRENCE,
             ]
 
         raise NotImplementedError(f"Invalid semantic type '{stype.value}'")
 
-    def compute(self, ser: Series) -> Any:
+    def compute(self, ser: Series, sep: Optional[str]) -> Any:
```
rusty1s, line 52:
```suggestion
    def compute(self, ser: Series, sep: Optional[str] = None) -> Any:
```

torch_frame/data/stats.py
```diff
@@ -59,21 +62,28 @@ def compute(self, ser: Series) -> Any:
...
+            ser = ser.explode().dropna()
+            count = ser.value_counts(ascending=False)
+            return count.index.tolist(), count.values.tolist()
         raise NotImplementedError(f"Invalid stat type '{self.value}'")
 
 
 def compute_col_stats(
     ser: Series,
     stype: torch_frame.stype,
+    sep: Optional[str],
```
rusty1s, line 79:
```suggestion
    sep: Optional[str] = None,
```

torch_frame/data/stats.py
yiweny: Renamed to `MULTI_COUNT`

torch_frame/data/mapper.py
yiweny: Removed `groupby`

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- Added `Occurrence` stat for `multicategorical` stype ([#151](https://github.com/pyg-team/pytorch-frame/pull/151))
```
weihua916, line 11: `MULTI_COUNT`

test/data/test_dataset.py
```diff
@@ -100,19 +100,21 @@ def test_converter():
     assert len(tf) == len(dataset)
 
 
-def test_multi_categorical_materialization():
-    data = {'a': ['A,B', 'B,C,A', '', 'B', None]}
+def test_multicategorical_materialization():
+    data = {'a': ['A|B', 'B|C|A', '', '', 'B', 'B|A', None]}
```
weihua916, line 104: nit 'a' --> 'multicat_col'

test/data/test_dataset.py
```diff
@@ -100,19 +100,21 @@ def test_converter():
...
     assert torch.equal(feat[0, 0], torch.tensor([1, 0], device=feat.device))
-    assert torch.equal(feat[1, 0], torch.tensor([0, 2, 1], device=feat.device))
-    assert feat[2, 0].numel() == 0
-    assert torch.equal(feat[4, 0], torch.tensor([-1], device=feat.device))
-    assert StatType.COUNT in dataset.col_stats['a']
-    assert dataset.col_stats['a'][StatType.COUNT][0] == ['B', 'A', 'C']
-    assert dataset.col_stats['a'][StatType.COUNT][1] == [3, 2, 1]
+    assert torch.equal(feat[1, 0], torch.tensor([0, 3, 1], device=feat.device))
+    assert torch.equal(feat[2, 0], torch.tensor([2], device=feat.device))
+    assert torch.equal(feat[6, 0], torch.tensor([-1], device=feat.device))
```
weihua916, line 112: Nit: Do you need `device=feat.device`?

test/data/test_dataset.py
weihua916, line 104: Mapping '' to a category is weird.

test/data/test_dataset.py
```diff
@@ -100,19 +100,21 @@ def test_converter():
...
+    feat = dataset.tensor_frame.feat_dict[stype.multicategorical]
     assert torch.equal(feat[0, 0], torch.tensor([1, 0], device=feat.device))
-    assert torch.equal(feat[1, 0], torch.tensor([0, 2, 1], device=feat.device))
-    assert feat[2, 0].numel() == 0
-    assert torch.equal(feat[4, 0], torch.tensor([-1], device=feat.device))
-    assert StatType.COUNT in dataset.col_stats['a']
-    assert dataset.col_stats['a'][StatType.COUNT][0] == ['B', 'A', 'C']
-    assert dataset.col_stats['a'][StatType.COUNT][1] == [3, 2, 1]
+    assert torch.equal(feat[1, 0], torch.tensor([0, 3, 1], device=feat.device))
+    assert torch.equal(feat[2, 0], torch.tensor([2], device=feat.device))
```
weihua916, line 111: IMO, this should map to torch.tensor([]).

test/data/test_mapper.py
```diff
@@ -39,11 +39,11 @@ def test_categorical_tensor_mapper():
...
 
-def test_multi_categorical_tensor_mapper():
+def test_multicategorical_tensor_mapper():
     ser = pd.Series(['A,B', 'B', '', 'C', 'B,C', None])
-    expected_values = torch.tensor([1, 0, 0, -1, 0, -1, -1])
-    expected_boundaries = torch.tensor([0, 2, 3, 3, 4, 6, 7])
-    mapper = MultiCategoricalTensorMapper(['B', 'A'], sep=",")
+    expected_values = torch.tensor([1, 0, 0, 2, 0, -1])
+    expected_boundaries = torch.tensor([0, 2, 3, 4, 4, 5, 6])
+    mapper = MultiCategoricalTensorMapper(['B', 'A', ''], sep=",")
```
weihua916, line 46: Similar here, mapping empty into a category is weird to me.

torch_frame/data/dataset.py
```diff
@@ -78,7 +80,7 @@ def __init__(
         col_to_stype: Dict[str, torch_frame.stype],
         col_stats: Dict[str, Dict[StatType, Any]],
         target_col: Optional[str] = None,
-        sep: str = ',',
+        sep: Optional[Dict[str, str]] = None,
```
weihua916, line 83: `col_to_sep`

torch_frame/data/dataset.py
```diff
@@ -116,9 +118,16 @@ def _get_mapper(self, col: str) -> TensorMapper:
...
             return CategoricalTensorMapper(index)
-        elif stype == torch_frame.multi_categorical:
-            index, _ = self.col_stats[col][StatType.COUNT]
-            return MultiCategoricalTensorMapper(index, sep=self.sep)
+        elif stype == torch_frame.multicategorical:
+            index, _ = self.col_stats[col][StatType.MULTI_COUNT]
+            if self.sep is None or col not in self.sep:
+                warnings.warn(
+                    "The separator is not specified for multicategorical"
+                    " column: {col}. ',' will be set as default separator.")
```
weihua916, line 126: Don't you need f-string?

torch_frame/data/dataset.py
```diff
@@ -116,9 +118,16 @@ def _get_mapper(self, col: str) -> TensorMapper:
...
+                if self.sep is None:
+                    self.sep = {}
+                self.sep[col] = ','
```
weihua916, line 129: Why do we set sep here? I think it's better to be set in init.

torch_frame/data/dataset.py
```diff
@@ -181,7 +188,7 @@ def __init__(
         col_to_stype: Dict[str, torch_frame.stype],
         target_col: Optional[str] = None,
         split_col: Optional[str] = None,
-        sep: str = ',',
+        sep: Dict[str, str] = {},
```
weihua916, line 191: `col_to_sep`
Also modify the doc-string.

torch_frame/data/dataset.py
```diff
@@ -335,15 +342,17 @@ def materialize(self, device: Optional[torch.device] = None,
...
+            if stype == torch_frame.multicategorical and (self.sep is None or
+                                                          col not in self.sep):
+                warnings.warn(
+                    "The separator is not specified for multicategorical"
+                    " column: {col}. ',' will be set as default separator.")
+                if self.sep is None:
+                    self.sep = {}
+                self.sep[col] = ','
+            self._col_stats[col] = compute_col_stats(
+                ser, stype, sep=self.sep.get(col, None))
```
weihua916, line 355: Remove this and put it in init

torch_frame/data/mapper.py
```diff
@@ -92,11 +92,12 @@ def backward(self, tensor: Tensor) -> pd.Series:
...
     r"""Maps any multi-categorical series into an index representation, with
     :obj:`-1` denoting missing values (NaN) and no value denoting not belonging
-    to any categories.
+    to any categories. If the cell contains unseen categories, it will be
+    ignored.
 
     Args:
         categories (List[Any]): A list of possible categories in the
-        multi-categorical column.
+        multi-categorical column sorted by occurrence.
```
weihua916, line 100: sorted by counts

torch_frame/data/dataset.py
weihua916: Also, do we need warning? Just logging seems enough to me.

torch_frame/data/mapper.py
```diff
@@ -129,13 +130,25 @@ def forward(
...
+        ser = ser.explode()
+        ser = pd.merge(
+            ser.rename('data'),
+            self.index.rename('index'),
+            how='left',
+            left_on='data',
+            right_index=True,
+        ).dropna()
+        ser['index'] = ser['index'].astype('int64')
+        values = torch.tensor(ser['index'].values, device=device)
```
weihua916, line 144: nit:
```
values = torch.tensor(ser['index'].values, device=device, dtype=torch.long)
```

torch_frame/data/mapper.py
```diff
@@ -129,13 +130,25 @@ def forward(
...
+        offset = ser.index.value_counts()
+        offset = offset.reindex(original_index, fill_value=0)
+        offset = pd.concat((pd.Series([0]), offset))
+        offset = torch.tensor(offset.values, device=device)
         offset = torch.cumsum(offset, dim=0)
```
weihua916, line 149: Just curious, why you don't need to cast to torch.long here?

torch_frame/data/dataset.py
weihua916: Add to doc-string the behavior of `None`

torch_frame/data/dataset.py
weihua916: Also, do we need warning if in the documentation, we clearly state that the default is ","?

weihua916:
For `col_to_sep`, I'd make it `Union[str, Dict[str, str]]`.
If type is `str`, we just use that for all multi-categorical columns. Otherwise, we use `sep` for each column.
So the default is `col_to_sep = ","`.

test/data/test_dataset.py
rusty1s: +1, these should be discarded.

torch_frame/__init__.py
```diff
@@ -16,10 +16,11 @@
     'numerical',
     'categorical',
     'text_embedded',
-    'multi_categorical',
+    'multicategorical',
     'TaskType',
     'NAStrategy',
     'TensorFrame',
+    'TensorData',
```
rusty1s, line 23: Is there a reason to expose it? I would prefer to keep it internal only.

torch_frame/data/dataset.py
```diff
@@ -116,9 +118,16 @@ def _get_mapper(self, col: str) -> TensorMapper:
...
-            index, _ = self.col_stats[col][StatType.COUNT]
-            return MultiCategoricalTensorMapper(index, sep=self.sep)
+        elif stype == torch_frame.multicategorical:
+            index, _ = self.col_stats[col][StatType.MULTI_COUNT]
+            if self.sep is None or col not in self.sep:
+                warnings.warn(
+                    "The separator is not specified for multicategorical"
+                    " column: {col}. ',' will be set as default separator.")
+                if self.sep is None:
+                    self.sep = {}
```
rusty1s, line 128: We can initialize this as empty as part of `__init__`: `self.sep = sep or {}`.

torch_frame/data/dataset.py
```diff
@@ -335,15 +342,17 @@ def materialize(self, device: Optional[torch.device] = None,
...
-                    lambda x: [cat.strip() for cat in x.split(self.sep)]
-                    if x is not None and x != '' else [])
-                ser = ser.explode()
-                ser = ser.dropna()
-            else:
-                ser = self.df[col]
-            self._col_stats[col] = compute_col_stats(ser, stype)
+            ser = self.df[col]
+            if stype == torch_frame.multicategorical and (self.sep is None or
+                                                          col not in self.sep):
```
rusty1s, line 347:
```suggestion
            if (stype == torch_frame.multicategorical and (self.sep is None or
                                                          col not in self.sep)):
```
will give you a better formatting. +1 no reason to check for `self.sep is None` if you do `self.sep = sep or {}` in constructor (sorry if this wasn't clear).

torch_frame/data/mapper.py
rusty1s: actually, we should always use `torch.from_numpy` instead for zero-copy.

rusty1s (commented):
Looks great, thanks for the updates <3. Should be good once comments are resolved.

torch_frame/data/dataset.py
weihua916: Also, I prefer making its type as either `str` or `Dict`, instead of `None`. If `str`, we use the shared separator across all columns.

test/data/test_dataset.py
yiweny: Removed

torch_frame/data/dataset.py
yiweny: I think it's good to have a warning for user.

torch_frame/data/mapper.py
yiweny: I think I obtained the `offset` from `ser.index.value_counts()`. So I assume `value_counts()` by default returns int64 type.

torch_frame/__init__.py
yiweny: Yeah, makes sense.

torch_frame/data/dataset.py
yiweny: How about `Optional[Union[str, Dict[str, str]]]`? So it can take either `None`, `str` or `Dict[str, str]`.

yiweny: @weihua916 If we are to default it to `,` anyways, can we set it as `Union[str, Dict[str, str]]` and defaults to `,`?

torch_frame/data/mapper.py
weihua916: Ok, I just wanted to confirm since you explicitly casted as `long` before

torch_frame/data/dataset.py
weihua916: I think we can follow https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html, where the default is set to `sep = ","`
It is weird to have `sep=None` also mean `sep = ","`. I think `sep=None` should mean automatic inference of the separator, but that's tricky to implement...

weihua916: > If we are to default it to , anyways, can we set it as Union[str, Dict[str, str]] and defaults to ","?

Yes, that makes most sense to me. Thanks

torch_frame/data/dataset.py
```diff
@@ -65,7 +67,10 @@ class DataFrameToTensorFrameConverter:
             column name into stats. Available as :obj:`dataset.col_stats`.
         target_col (str, optional): The column used as target.
             (default: :obj:`None`)
-        sep (str): Separator for multi-categorical columns. (default: :obj:`,`)
+        col_to_sep (Union[str, Dict[str, str]], optional): A dictionary that
+            maps each multi_categorical column to its delimiter or a string
+            indicating the delimiter for all multi_categorical columns. If not
+            specified, then :obj:`,` will be used. (default: :obj:`None`)
```
weihua916, line 73: Update the description and disallow `optional`.
If `str`, then we will use that as a separator across columns.

torch_frame/data/dataset.py
```diff
@@ -78,13 +83,24 @@ def __init__(
         col_to_stype: Dict[str, torch_frame.stype],
         col_stats: Dict[str, Dict[StatType, Any]],
         target_col: Optional[str] = None,
-        sep: str = ',',
+        col_to_sep: Optional[Union[str, Dict[str, str]]] = None,
```
weihua916, line 86: same here.
`col_to_sep = ','`

torch_frame/data/dataset.py
weihua916: Could you update this throughout the code? Thanks.

torch_frame/data/dataset.py
```diff
@@ -78,13 +83,24 @@ def __init__(
...
+        if isinstance(col_to_sep, Dict):
+            for col in col_to_stype:
+                if (col_to_stype[col] == torch_frame.multicategorical
+                        and col not in col_to_sep):
+                    warnings.warn(
+                        "The delimiter is not specified for multicategorical "
+                        f"column: {col}. ',' will be set as default delimiter."
+                    )
+
+                    col_to_sep[col] = ','
```
weihua916, line 99: Update the logic and remove a warning.

torch_frame/data/mapper.py
weihua916: +1

torch_frame/data/dataset.py
yiweny: Updated~

torch_frame/data/dataset.py
yiweny: Updated~

weihua916 (approved):
Looks good, thanks. Made some further fixes and code refactoring.
Feel free to merge it once you've confirmed my fixes are all alright! Otherwise, lmk if you have any concerns.

test/data/test_dataset.py
```diff
@@ -100,19 +101,23 @@ def test_converter():
...
-    assert torch.equal(feat[1, 0], torch.tensor([0, 2, 1], device=feat.device))
-    assert feat[2, 0].numel() == 0
-    assert torch.equal(feat[4, 0], torch.tensor([-1], device=feat.device))
-    assert StatType.COUNT in dataset.col_stats['a']
-    assert dataset.col_stats['a'][StatType.COUNT][0] == ['B', 'A', 'C']
-    assert dataset.col_stats['a'][StatType.COUNT][1] == [3, 2, 1]
+    feat = dataset.tensor_frame.feat_dict[stype.multicategorical]
+    assert torch.equal(feat[0, 0], torch.tensor([1, 0]))
+    assert torch.equal(feat[1, 0], torch.tensor([0, 2, 1]))
+    assert torch.equal(feat[6, 0], torch.tensor([-1]))
```
weihua916, line 113: Can we add tests for 2, 3, 4, 5 as well?

yiweny: While I was adding tests, I realized that I didn't account for the case where some categories can appear multiple times. I added some code and fixed test cases to accommodate that case.

weihua916: Good catch. thanks

PR #155, opened 2023-10-29
Add multi categorical encoder

yiweny:


torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -194,6 +195,65 @@ def encode_forward(self, feat: Tensor) -> Tensor:
...
+    """
+    supported_stypes = {stype.multicategorical}
+
+    def __init__(
+        self,
+        out_channels: Optional[int] = None,
+        stats_list: Optional[List[Dict[StatType, Any]]] = None,
+        stype: Optional[stype] = None,
+        post_module: Optional[Module] = None,
+        na_strategy: Optional[NAStrategy] = None,
```
rusty1s, line 216: I assume we would need to error out in case `na_strategy` is set to something else than `NAStrategy.ZERO`.

rusty1s (approved):

torch_frame/nn/encoder/stype_encoder.py
yiweny: Good point.

yiweny: Acutally, `NAStrategy.ZERO` is the same as `NAStrategy.MOST_FREQUENT` because the most frequent class will always be mapped to 0.

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -194,6 +200,67 @@ def encode_forward(self, feat: Tensor) -> Tensor:
...
 
 
+class MultiCategoricalEmbeddingEncoder(StypeEncoder):
+    r"""An embedding look-up based encoder for multi_categorical features. It
+    applies :class:`torch.nn.EmbeddingBag` for each categorical feature and
+    concatenates the output embeddings.
+
+    Args:
+        out_channels (int, optional): Size of each embedding vector.
+            (default: :obj:`None`)
```
weihua916, line 210: nit: You can remove `out_channels`. This is LAZY_ATTR shared `StypeEncoder`.

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -194,6 +200,67 @@ def encode_forward(self, feat: Tensor) -> Tensor:
...
+                EmbeddingBag(num_categories + 1, self.out_channels,
+                             padding_idx=0, mode=self.mode))
+        self.reset_parameters()
+
+    def reset_parameters(self):
+        super().reset_parameters()
+        for emb in self.embs:
+            emb.reset_parameters()
+
+    def encode_forward(self, feat: TensorData) -> Tensor:
```
weihua916, line 249:
```python
feat: MultiNestedTensor
```
since we only allow `MultiNestedTensor` in multi-categorical.

weihua916 (approved):
Very clean, thanks!

torch_frame/datasets/fake.py
```diff
@@ -69,16 +69,13 @@ def __init__(
...
-            # multi-categorical columns when it's added.
             for col_name in ['mult_1']:
                 arr = np.random.randint(0, 3, size=(num_rows, 2))
                 arr = arr.astype(str)
                 arr = np.apply_along_axis(lambda x: ','.join(x), 1, arr)
-                if with_nan:
-                    arr[1::2] = None
                 df_dict[col_name] = list(arr)
+                if with_nan:
+                    df_dict[col_name][0] = None
```
weihua916, line 81: df_dict[col_name][1::2] = None

weihua916 (approved):

torch_frame/nn/encoder/stype_encoder.py
rusty1s: This is not true, right? `ZERO` would produce a zero embedding, while `MOST_FREQUENT` should pick up the embedding from the most frequent category.

PR #157, opened 2023-10-29
Add basic text tokenized

zechengz:
Add basic text tokenized functionalities. Encoder change to be in the following PR.

### DIFF
--- a/test/datasets/test_fake_dataset.py
+++ b/test/datasets/test_fake_dataset.py
@@ -1,10 +1,15 @@
+from typing import Dict
+
 import pytest
 import torch
 
 import torch_frame
 from torch_frame.config.text_embedder import TextEmbedderConfig
+from torch_frame.config.text_tokenizer import TextTokenizerConfig
+from torch_frame.data import MultiNestedTensor
 from torch_frame.datasets import FakeDataset
 from torch_frame.testing.text_embedder import HashTextEmbedder
+from torch_frame.testing.text_tokenizer import WhiteSpaceTokenizer
 
 
 @pytest.mark.parametrize('with_nan', [True, False])
@@ -18,13 +23,19 @@ def test_fake_dataset(with_nan):
             torch_frame.numerical,
             torch_frame.categorical,
             torch_frame.text_embedded,
+            torch_frame.text_tokenized,
         ],
         text_embedder_cfg=TextEmbedderConfig(
             text_embedder=HashTextEmbedder(out_channels), batch_size=None),
+        text_tokenizer_cfg=TextTokenizerConfig(
+            text_tokenizer=WhiteSpaceTokenizer(num_hash_bins=64),
+            batch_size=2),
     )
     assert str(dataset) == 'FakeDataset()'
     assert len(dataset) == num_rows
-    assert dataset.feat_cols == ['a', 'b', 'c', 'x', 'y', 'text_1', 'text_2']
+    assert dataset.feat_cols == [
+        'a', 'b', 'c', 'x', 'y', 'text_1', 'text_2', 'text_3', 'text_4'
+    ]
     assert dataset.target_col == 'target'
 
     dataset = dataset.materialize()
@@ -50,3 +61,9 @@ def test_fake_dataset(with_nan):
     assert feat_text_embedded.shape == (
         num_rows, len(tensor_frame.col_names_dict[torch_frame.text_embedded]),
         out_channels)
+
+    feat_text_tokenized: Dict[str, MultiNestedTensor] = tensor_frame.feat_dict[
+        torch_frame.text_tokenized]
+    assert feat_text_tokenized['input_ids'].dtype == torch.long
+    assert feat_text_tokenized['input_ids'].num_rows == num_rows
+    assert feat_text_tokenized['input_ids'].num_cols == 2
--- a/test/datasets/test_fake_dataset.py
+++ b/test/datasets/test_fake_dataset.py
@@ -9,7 +9,7 @@ from torch_frame.config.text_tokenizer import TextTokenizerConfig
 from torch_frame.data import MultiNestedTensor
 from torch_frame.datasets import FakeDataset
 from torch_frame.testing.text_embedder import HashTextEmbedder
-from torch_frame.testing.text_tokenizer import WhiteSpaceTokenizer
+from torch_frame.testing.text_tokenizer import WhiteSpaceHashTokenizer
 
 
 @pytest.mark.parametrize('with_nan', [True, False])
@@ -28,7 +28,7 @@ def test_fake_dataset(with_nan):
         text_embedder_cfg=TextEmbedderConfig(
             text_embedder=HashTextEmbedder(out_channels), batch_size=None),
         text_tokenizer_cfg=TextTokenizerConfig(
-            text_tokenizer=WhiteSpaceTokenizer(num_hash_bins=64),
+            text_tokenizer=WhiteSpaceHashTokenizer(num_hash_bins=64),
             batch_size=2),
     )
     assert str(dataset) == 'FakeDataset()'
--- a/test/datasets/test_fake_dataset.py
+++ b/test/datasets/test_fake_dataset.py
@@ -1,15 +1,10 @@
-from typing import Dict
-
 import pytest
 import torch
 
 import torch_frame
 from torch_frame.config.text_embedder import TextEmbedderConfig
-from torch_frame.config.text_tokenizer import TextTokenizerConfig
-from torch_frame.data import MultiNestedTensor
 from torch_frame.datasets import FakeDataset
 from torch_frame.testing.text_embedder import HashTextEmbedder
-from torch_frame.testing.text_tokenizer import WhiteSpaceHashTokenizer
 
 
 @pytest.mark.parametrize('with_nan', [True, False])
@@ -23,19 +18,13 @@ def test_fake_dataset(with_nan):
             torch_frame.numerical,
             torch_frame.categorical,
             torch_frame.text_embedded,
-            torch_frame.text_tokenized,
         ],
         text_embedder_cfg=TextEmbedderConfig(
             text_embedder=HashTextEmbedder(out_channels), batch_size=None),
-        text_tokenizer_cfg=TextTokenizerConfig(
-            text_tokenizer=WhiteSpaceHashTokenizer(num_hash_bins=64),
-            batch_size=2),
     )
     assert str(dataset) == 'FakeDataset()'
     assert len(dataset) == num_rows
-    assert dataset.feat_cols == [
-        'a', 'b', 'c', 'x', 'y', 'text_1', 'text_2', 'text_3', 'text_4'
-    ]
+    assert dataset.feat_cols == ['a', 'b', 'c', 'x', 'y', 'text_1', 'text_2']
     assert dataset.target_col == 'target'
 
     dataset = dataset.materialize()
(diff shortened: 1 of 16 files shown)


test/datasets/test_fake_dataset.py
```diff
@@ -50,3 +61,9 @@ def test_fake_dataset(with_nan):
     assert feat_text_embedded.shape == (
         num_rows, len(tensor_frame.col_names_dict[torch_frame.text_embedded]),
         out_channels)
+
+    feat_text_tokenized: Dict[str, MultiNestedTensor] = tensor_frame.feat_dict[
```
yiweny, line 65:
```suggestion
    feat_text_tokenized: TensorData = tensor_frame.feat_dict[
```

torch_frame/config/text_tokenizer.py
```diff
@@ -0,0 +1,22 @@
...
+    Args:
+        text_tokenizer (callable): A callable text tokenizer that takes a list
+            of strings as input and output the dictionary of
+            :class:`MultiNestedTensor`.
+        batch_size (int, optional): Batch size to use when encoding the
+            sentences. If set to :obj:`None`, the text embeddings will
+            be obtained in a full-batch manner. (default: :obj:`None`)
+
+    """
+    text_tokenizer: Callable[[List[str]], Dict[str, MultiNestedTensor]]
```
yiweny, line 21:
```suggestion
    text_tokenizer: Callable[[List[str]], TensorData]
```

torch_frame/data/tensor_frame.py
```diff
@@ -83,6 +83,11 @@ def validate(self):
             num_cols = len(self.col_names_dict[stype_name])
             if num_cols == 0:
                 empty_stypes.append(stype_name)
+
+            # Get any value if feat is a dictionary
+            if isinstance(feat, dict):
+                feat = next(iter(feat.values()))
```
yiweny, line 89: It's better to iterate the keys and check for all values.

torch_frame/data/dataset.py
```diff
@@ -184,11 +203,19 @@ def __call__(
...
-        feat_dict = {
-            stype: (torch.stack(xs, dim=1) if not stype.use_multi_nested_tensor
-                    else MultiNestedTensor.stack(xs, dim=1))
-            for stype, xs in xs_dict.items()
-        }
+
+        feat_dict = {}
+        for stype, xs in xs_dict.items():
+            if stype.use_multi_nested_tensor:
+                feat_dict[stype] = MultiNestedTensor.stack(xs, dim=1)
```
yiweny, line 210:
```suggestion
                feat_dict[stype] = MultiNestedTensor.cat(xs, dim=1)
```

torch_frame/data/mapper.py
```diff
@@ -209,3 +209,59 @@ def forward(
...
+
+
+class TextTokenizationTensorMapper(TensorMapper):
+    r"""Tokenize any text series into a dictionary of
+    :class:`MultiNestedTensor`.
+
+    Args:
+        text_tokenizer (callable): A callable function that takes list of
+            strings and returns a dictionary of :class:`MultiNestedTensor`.
+            The values of the dictionary contains for example tokens and
```
yiweny, line 221: Correct me if I am wrong, but I think the set of keys should be the same for all tokenizer, i.e. ids, attention_masks and tokens. Can you be more specific in the documentation.

torch_frame/data/mapper.py
```diff
@@ -209,3 +209,59 @@ def forward(
...
+            attention masks.
+        batch_size (int, optional): The mini-batch size used for the text
```
yiweny, line 223: Just curious, why is mini-batch necessary if we are not having parallelism

torch_frame/data/multi_nested_tensor.py
```diff
@@ -28,7 +28,6 @@ def __init__(
     ):
         assert offset[0] == 0
         assert offset[-1] == len(values)
-        assert len(offset) == num_rows * num_cols + 1
```
yiweny, line 31: Probably good to keep this?

torch_frame/data/multi_nested_tensor.py
```diff
@@ -224,6 +231,29 @@ def stack(xs: List['MultiNestedTensor'],
         else:
             raise NotImplementedError
 
+    @staticmethod
+    def cat(xs: List['MultiNestedTensor'],
```
yiweny, line 235: I think in PyTorch `Tuple` of `Tensors` are also supported. We should implement this function with the same arguments as PyTorch `cat`.

torch_frame/data/multi_nested_tensor.py
```diff
@@ -224,6 +231,29 @@ def stack(xs: List['MultiNestedTensor'],
...
+            num_cols = sum(x.num_cols for x in xs)
+        else:
+            raise NotImplementedError
+        values = torch.cat([x.values for x in xs], dim=0)
+        accum_offset = 0
+        offset = []
+        for x in xs:
+            offset.append(x.offset + accum_offset)
+            accum_offset = accum_offset + x.offset[-1]
+        offset = torch.cat(offset, dim=0)
```
yiweny, line 253: `x.offset` contains a 0 index. It won't work with this logic because a dummy 0 will be added for each `x` in `xs`.

yiweny (changes requested):
I think you need to add a test case for `cat` function in `MultiNestedTensor`. Also there is some issue with the logic.

torch_frame/data/multi_nested_tensor.py
```diff
@@ -224,6 +231,29 @@ def stack(xs: List['MultiNestedTensor'],
...
+            num_rows = xs[0].num_rows
+            num_cols = sum(x.num_cols for x in xs)
+        else:
+            raise NotImplementedError
+        values = torch.cat([x.values for x in xs], dim=0)
+        accum_offset = 0
+        offset = []
+        for x in xs:
+            offset.append(x.offset + accum_offset)
+            accum_offset = accum_offset + x.offset[-1]
```
weihua916, line 252: This logic is only applicable to `dim=0`.

weihua916:
Thanks @zechengz As @yiweny said, it'd be great if you could separate `cat` method out and thoroughly test it in both `dim=0` and `dim=1`.

torch_frame/data/mapper.py
zechengz: No, they are not the same. For example BERT has attention mask, token input ids and sequence id. But roberta only has attention mask and token input ids. Will update the docs.

torch_frame/data/mapper.py
zechengz: Tokenizer does not have this issue as they don't need GPU. For convenience I just keep the interface the same. @weihua916 Do you think we can remove the `batch_size` for text tokenized?

torch_frame/data/multi_nested_tensor.py
zechengz: Right, moved to `dim=0`

torch_frame/data/multi_nested_tensor.py
zechengz: Right, will be fixed in the following PR

torch_frame/data/mapper.py
zechengz: Discussed with @weihua916 , for now just keep it same with text embedder cfg. In the future we can easily delete it.

torch_frame/data/multi_nested_tensor.py
```diff
@@ -358,6 +358,23 @@ def stack(xs: List['MultiNestedTensor'],
...
+            return xs[0]
+
+        if dim < 0:
+            dim += xs[0].ndim
+
+        if dim == 0:
+            raise NotImplementedError
+        elif dim == 1:
+            raise NotImplementedError
+        else:
```
yiweny, line 375: Do we need to support cat on the third dimension? Maybe raise a different error here.

torch_frame/data/mapper.py
```diff
@@ -209,3 +209,60 @@ def forward(
...
+        device: Optional[torch.device] = None,
+    ) -> Dict[str, MultiNestedTensor]:
+        ser = ser.astype(str)
+        ser_list = ser.tolist()
+
+        seq_tokens_dict: Optional[Dict[str, MultiNestedTensor]] = None
+        if self.batch_size is None:
+            seq_tokens_dict = self.text_tokenizer(ser_list)
+            for key in seq_tokens_dict:
+                seq_tokens_dict[key].to(device)
```
weihua916, line 250:
```python
for key, val in seq_tokens_dict:
    seq_tokens_dict[key] = value.to(device)
```

yiweny (approved):

torch_frame/data/tensor_frame.py
```diff
@@ -83,20 +83,42 @@ def validate(self):
...
+                            f"feature is {num_cols}, which does not align "
+                            f"with the column dimensionality of "
+                            f"feat_dict['{stype_name}']['{key}'] "
+                            f"(got {feat.size(1)})")
+                    if feat.size(0) != num_rows:
+                        raise ValueError(
+                            f"The length of elements in "
+                            f"feat_dict['{stype_name}']['{key}']"
+                            f"is not aligned, got {feat.size(0)} but "
+                            f"expected {num_rows}.")
```
weihua916, line 121: This is highly repetitive... Can we somehow make this cleaner?

torch_frame/testing/text_tokenizer.py
```diff
@@ -0,0 +1,33 @@
...
+                 device: Optional[torch.device] = None):
+        self.device = device
+        self.num_hash_bins = num_hash_bins
+
+    def __call__(self, sentences: List[str]) -> Dict[str, MultiNestedTensor]:
+        res = []
+        for s in sentences:
+            tokens = s.split(' ')
+            idx = torch.LongTensor(
+                [hash(t) % self.num_hash_bins for t in tokens])
```
weihua916, line 29: torch.tensor

torch_frame/data/mapper.py
```diff
@@ -209,3 +209,60 @@ def forward(
...
+    def forward(
+        self,
+        ser: Series,
+        *,
+        device: Optional[torch.device] = None,
+    ) -> Dict[str, MultiNestedTensor]:
+        ser = ser.astype(str)
+        ser_list = ser.tolist()
+
+        seq_tokens_dict: Optional[Dict[str, MultiNestedTensor]] = None
```
weihua916, line 246: weird to set it to None at the beggining.

torch_frame/data/mapper.py
```diff
@@ -209,3 +209,60 @@ def forward(
...
+            if seq_tokens_dict is None or len(seq_tokens_dict) == 0:
+                seq_tokens_dict = batch_seq_tokens_dict
+            else:
+                for key in seq_tokens_dict:
+                    seq_tokens_dict[key] = MultiNestedTensor.cat(
+                        [seq_tokens_dict[key], batch_seq_tokens_dict[key]],
+                        dim=0)
+        for key in seq_tokens_dict:
+            seq_tokens_dict[key].to(device)
+        return seq_tokens_dict
```
weihua916, line 265: re-write so that we do concatenation at last.

torch_frame/data/multi_nested_tensor.py
weihua916: We should raise an error.

weihua916 (commented):
Thanks, although I found this PR is severely untested.

From next time, we should implement things sequentially and test PR more thoroughly (first implement `cat(dim=1)`, then this functionality).

torch_frame/data/mapper.py
weihua916: same for other places.

torch_frame/data/tensor_frame.py
weihua916: You can make this as a function block and call it twice.
```python
if feat.dim() < 2:
                        raise ValueError(f"feat_dict['{stype_name}']['{key}'] "
                                         f"must be at least 2-dimensional")
                    if num_cols != feat.size(1):
                        raise ValueError(
                            f"The expected number of columns for {stype_name} "
                            f"feature is {num_cols}, which does not align "
                            f"with the column dimensionality of "
                            f"feat_dict['{stype_name}']['{key}'] "
                            f"(got {feat.size(1)})")
                    if feat.size(0) != num_rows:
                        raise ValueError(
                            f"The length of elements in "
                            f"feat_dict['{stype_name}']['{key}']"
                            f"is not aligned, got {feat.size(0)} but "
                            f"expected {num_rows}.")
```

PR #161, opened 2023-10-31
Supports advanced indexing of MultiNestedTensor (part 2)

weihua916:
Add concat dim=0 for `MultiNestedTensor`

Next PRs
- Column slicing
- Column concat

### DIFF
--- a/torch_frame/data/multi_nested_tensor.py
+++ b/torch_frame/data/multi_nested_tensor.py
@@ -353,6 +353,8 @@ class MultiNestedTensor:
     def cat(xs: Union[Tuple['MultiNestedTensor', ...],
                       List['MultiNestedTensor']],
             dim: int = 0) -> 'MultiNestedTensor':
+        if len(xs) == 0:
+            raise RuntimeError("Cannot concatenate a list of length 0.")
         assert isinstance(xs[0], MultiNestedTensor)
         if dim == 0 or dim + xs[0].ndim == 0:
             num_rows = sum(x.num_rows for x in xs)
--- a/torch_frame/data/multi_nested_tensor.py
+++ b/torch_frame/data/multi_nested_tensor.py
@@ -371,10 +371,13 @@ class MultiNestedTensor:
             return MultiNestedTensor(num_rows=num_rows, num_cols=num_cols,
                                      values=values, offset=offset)
         elif dim == 1 or dim == -2:
-            # TODO Weihua implement this
-            return xs[0]
+            if len(xs) == 1:
+                return xs[0]
+            else:
+                # TODO Weihua implement this
+                raise NotImplementedError
         else:
-            raise RuntimeError(f"Unsupported dim={dim} for slice.")
+            raise RuntimeError(f"Unsupported dim={dim} for concat.")
 
     def select(
         self,
--- a/torch_frame/data/multi_nested_tensor.py
+++ b/torch_frame/data/multi_nested_tensor.py
@@ -223,7 +223,7 @@ class MultiNestedTensor:
 
     def single_index_select(self, index: int, dim: int) -> 'MultiNestedTensor':
         r"""Get :obj:`index`-th row (:obj:`dim=0`) or column (:obj:`dim=1`)"""
-        if dim == 0 or dim + self.ndim + self.ndim == 0:
+        if dim == 0 or dim + self.ndim == 0:
             index = self._to_positive_index(index, dim=0)
             start_idx = index * self.num_cols
             end_idx = (index + 1) * self.num_cols + 1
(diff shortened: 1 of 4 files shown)


torch_frame/data/multi_nested_tensor.py
```diff
@@ -350,13 +350,34 @@ def dtype(self) -> torch.dtype:
...
     # Static methods ##########################################################
     @staticmethod
-    def stack(xs: List['MultiNestedTensor'],
-              dim: int = 0) -> 'MultiNestedTensor':
-        # TODO: To be implemented.
-        if len(xs) == 1:
-            return xs[0]
+    def cat(xs: List['MultiNestedTensor'],
+            dim: int = 0) -> 'MultiNestedTensor':
+        if dim == 0 or dim == -3:
```
yiweny, line 355: May be do something like dim = dim + MultiNestedTensor.dim if dim < 0 else dim rather than having `dim == 0 or dim == -3:`

torch_frame/data/multi_nested_tensor.py
```diff
@@ -350,13 +350,34 @@ def dtype(self) -> torch.dtype:
...
+                    raise RuntimeError(
+                        "num_cols must be the same across a list of input "
+                        "multi nested tensors.")
+            values = torch.cat([x.values for x in xs], dim=0)
+            accum_offset = 0
+            offset = []
+            for x in xs[:-1]:
+                offset.append(x.offset[:-1] + accum_offset)
+                accum_offset = accum_offset + x.offset[-1]
+            offset.append(xs[-1].offset + accum_offset)
```
yiweny, line 369: This operation is O(num_cols * num_rows) and then you then need to do `torch.cat` which is also quadratic. It would be inefficient. 
Can we do something like 
```
offset = torch.zeros(num_rows * num_cols + 1)
for i in range(1, num_rows * num_cols + 1):
    offset[i] = ...
```
So we can half the runtime?

torch_frame/data/multi_nested_tensor.py
```diff
@@ -350,13 +350,34 @@ def dtype(self) -> torch.dtype:
 
     # Static methods ##########################################################
     @staticmethod
-    def stack(xs: List['MultiNestedTensor'],
-              dim: int = 0) -> 'MultiNestedTensor':
-        # TODO: To be implemented.
-        if len(xs) == 1:
-            return xs[0]
+    def cat(xs: List['MultiNestedTensor'],
```
yiweny, line 353: Should we support tuple here as well?

torch_frame/data/multi_nested_tensor.py
weihua916: torch.cat is not quadratic. I think this is ok.

torch_frame/data/multi_nested_tensor.py
weihua916: That'll require two lines. I think current one is simple enough.

torch_frame/data/multi_nested_tensor.py
weihua916: Yes, good point. Added and tested.

torch_frame/data/multi_nested_tensor.py
weihua916: Addressed.

torch_frame/data/multi_nested_tensor.py
weihua916: Hmm, I think this usage of torch.cat is pretty standard to me.

yiweny (approved):

torch_frame/data/multi_nested_tensor.py
```diff
@@ -350,13 +350,36 @@ def dtype(self) -> torch.dtype:
...
-        if len(xs) == 1:
-            return xs[0]
+    def cat(xs: Union[Tuple['MultiNestedTensor', ...],
+                      List['MultiNestedTensor']],
+            dim: int = 0) -> 'MultiNestedTensor':
+        assert isinstance(xs[0], MultiNestedTensor)
+        if dim == 0 or dim + xs[0].ndim == 0:
+            num_rows = sum(x.num_rows for x in xs)
+            num_cols = xs[0].num_cols
+            for x in xs:
```
zechengz, line 360: nit
```suggestion
            for x in xs[1:]:
```

zechengz (approved):

torch_frame/data/multi_nested_tensor.py
weihua916: thanks.

torch_frame/data/multi_nested_tensor.py
weihua916: Ok, I managed to make `cat` more memory efficient: https://github.com/pyg-team/pytorch-frame/pull/161/commits/aff3240c9b3416d26edb4a445d868f37c83a3bbc

PR #168, opened 2023-11-01
add stats for sequence_numerical and some more tests to multicategorical stypes

yiweny:


torch_frame/data/stats.py
```diff
@@ -37,19 +37,27 @@ def stats_for_stype(stype: torch_frame.stype) -> List['StatType']:
...
+                StatType.STD,
+                StatType.QUANTILES,
+            ]
         }
         return stats_type.get(stype, [])
 
     def compute(self, ser: Series, sep: Optional[str] = None) -> Any:
         if self == StatType.MEAN:
-            return np.mean(ser.values).item()
+            flattened = np.hstack(np.hstack(ser.values))
```
yiweny, line 51: Not sure if this is the best method, but it works with both `sequence_numerical` and `numerical`. 
If the `ser` represents `sequence_numerical`, 
then `ser.values` would look like `array([[list([1, 3.0])], [list([1, 2, 3.0])]], dtype=object)`, so we need to hstack twice.

test/datasets/test_fake_dataset.py
```diff
@@ -17,14 +18,18 @@ def test_fake_dataset(with_nan):
...
             torch_frame.text_embedded,
         ],
         text_embedder_cfg=TextEmbedderConfig(
             text_embedder=HashTextEmbedder(out_channels), batch_size=None),
     )
     assert str(dataset) == 'FakeDataset()'
     assert len(dataset) == num_rows
-    assert dataset.feat_cols == ['a', 'b', 'c', 'x', 'y', 'text_1', 'text_2']
+    assert dataset.feat_cols == [
+        'a', 'b', 'c', 'x', 'y', 'mult_1', 'seq_1', 'seq_2', 'text_1', 'text_2'
```
weihua916, line 31: rename to `seq_num_1`, `seq_num_2` to accomodate `seq_cat` later.

torch_frame/data/stats.py
weihua916: Looks good to me.

torch_frame/datasets/fake.py
```diff
@@ -80,6 +80,16 @@ def __init__(
...
                 col_to_stype[col_name] = stype.multicategorical
+        if stype.sequence_numerical in stypes:
+            for col_name in ['seq_1', 'seq_2']:
+                arr = []
+                for i in range(num_rows):
+                    list_len = np.random.randint(1, 3)
+                    arr.append(np.random.randint(0, 3, size=list_len).tolist())
+                df_dict[col_name] = arr
+                if with_nan:
+                    df_dict[col_name][0] = None
```
weihua916, line 91: can we add nan to the element of list? like `[0.2, nan, 0.7]`

torch_frame/datasets/fake.py
```diff
@@ -80,6 +80,16 @@ def __init__(
                 if with_nan:
                     df_dict[col_name][0] = None
                 col_to_stype[col_name] = stype.multicategorical
+        if stype.sequence_numerical in stypes:
+            for col_name in ['seq_1', 'seq_2']:
+                arr = []
+                for i in range(num_rows):
+                    list_len = np.random.randint(1, 3)
+                    arr.append(np.random.randint(0, 3, size=list_len).tolist())
```
weihua916, line 88: super nit: I feel you don't need `numpy` here.
Just use pure Python. something like
```python
sequence_length = random.randint(5, 20)
sequence = [random.randint(1, 100) for _ in range(sequence_length)]
```

weihua916 (approved):
Thanks! Added some minor comments. LGTM after addressing them.

PR #174, opened 2023-11-05
Add text tokenized functionality and update text examples

zechengz:
- Enable text tokenized
  - Add text tokenized encoder
  - Fix several bugs
- Update the text example 
  - Finetuning text model using LoRA (`peft`), which has better performance than finetuning the text model directly and also better than using `text_embedded`
  - Add mean pooling in the example as suggested in https://arxiv.org/pdf/1908.10084.pdf
  - Remove `sentence-transformers` and use all text tokenizer and model from `transformers`


torch_frame/data/dataset.py
```diff
@@ -551,10 +551,13 @@ def get_split(self, split: str) -> 'Dataset':
...
+        train_split = self.get_split('train')
+        val_split = self.get_split('val')
+        test_split = self.get_split('test')
         return (
-            self.get_split('train'),
-            self.get_split('val'),
-            self.get_split('test'),
+            train_split,
+            val_split,
+            test_split,
```
yiweny, line 560: Can you remove this change as it seems irrelevant to this pr?

torch_frame/data/tensor_frame.py
```diff
@@ -207,13 +236,37 @@ def __copy__(self) -> 'TensorFrame':
...
+            else:
+                x = x.to(*args, **kwargs)
+            return x
+
+        return self._apply(fn)
 
     def cpu(self, *args, **kwargs):
-        return self._apply(lambda x: x.cpu(*args, **kwargs))
+        def fn(x):
+            if isinstance(x, dict):
```
yiweny, line 251:
```suggestion
            if isinstance(x, Dict):
```

torch_frame/data/tensor_frame.py
```diff
@@ -207,13 +236,37 @@ def __copy__(self) -> 'TensorFrame':
     # Device Transfer #########################################################
 
     def to(self, *args, **kwargs):
-        return self._apply(lambda x: x.to(*args, **kwargs))
+        def fn(x):
+            if isinstance(x, dict):
```
yiweny, line 240:
```suggestion
            if isinstance(x, Dict):
```

torch_frame/data/tensor_frame.py
```diff
@@ -207,13 +236,37 @@ def __copy__(self) -> 'TensorFrame':
...
+            else:
+                x = x.cpu(*args, **kwargs)
+            return x
+
+        return self._apply(fn)
 
     def cuda(self, *args, **kwargs):
-        return self._apply(lambda x: x.cuda(*args, **kwargs))
+        def fn(x):
+            if isinstance(x, dict):
```
yiweny, line 262:
```suggestion
            if isinstance(x, Dict):
```

torch_frame/data/tensor_frame.py
```diff
@@ -207,13 +236,37 @@ def __copy__(self) -> 'TensorFrame':
...
+                for key in x:
+                    x[key] = x[key].cuda(*args, **kwargs)
+            else:
+                x = x.cuda(*args, **kwargs)
+            return x
+
+        return self._apply(fn)
```
yiweny, line 269: The logic seems to be repetitive. Can you just call `to('cuda')` here? Same with `to('cpu')`

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -590,3 +594,71 @@ def encode_forward(self, feat: Tensor) -> Tensor:
...
+        stype: Optional[stype] = None,
+        post_module: Optional[Module] = None,
+        na_strategy: Optional[NAStrategy] = None,
+        in_channels: Optional[int] = None,
+        model: Optional[torch.nn.Module] = None,
+    ):
+        if in_channels is None:
+            raise ValueError("Please manually specify the `in_channels`, "
+                             "which is the text embedding dimensionality.")
+        if model is None:
```
yiweny, line 631: Why not make `model` not optional?

torch_frame/data/mapper.py
```diff
@@ -291,7 +291,15 @@ def forward(
...
-                xs = [[item[key]] for item in tokenized_list]
+                xs = []
+                for item in tokenized_list:
+                    if item[key].ndim == 1:
+                        xs.append([item[key]])
+                    elif item[key].ndim == 2:
+                        xs.append([row for row in item[key]])
+                    else:
+                        raise ValueError(f'{key} has `ndim` not '
+                                         f'equal to 1 or 2.')
```
yiweny, line 302: Do you mind explaining the logic here a little bit? Thanks~

test/datasets/test_fake_dataset.py
```diff
@@ -88,3 +89,9 @@ def test_fake_dataset(with_nan):
     assert isinstance(feat_text_tokenized['input_ids'], MultiNestedTensor)
     assert feat_text_tokenized['input_ids'].dtype == torch.int64
     assert feat_text_tokenized['input_ids'].shape == (num_rows, 2, -1)
+
+    # Test dataset split
+    train_dataset, val_dataset, test_dataset = dataset.split()
+    assert train_dataset.num_rows == num_rows - 2
+    assert val_dataset.num_rows == 1
+    assert test_dataset.num_rows == 1
```
yiweny, line 97: Mind doing this in a separate pr? This piece of code seems to be unrelated to text tokenizer.

yiweny (approved):
Thank you for the examples. I have some minor questions and comments. 
In addition, it think it might be better to put text embedding and text tokenizer in different examples.

examples/ft_transformer_text.py
```diff
@@ -51,21 +52,143 @@ def __call__(self, sentences: List[str]) -> Tensor:
 parser.add_argument('--lr', type=float, default=0.0001)
 parser.add_argument('--epochs', type=int, default=100)
 parser.add_argument('--seed', type=int, default=0)
+parser.add_argument('--finetune', action='store_true')
+parser.add_argument('--lora', action='store_true')
```
vid-koci, line 56: Can we add a warning/error/comment if somebody activates lora but not fine-tune? In that case lora flag is useless.

torch_frame/data/tensor_frame.py
vid-koci: `Dict` is for type annotation only, I believe. I don't think that this suggestion will achieve the intended effect

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -590,3 +594,71 @@ def encode_forward(self, feat: Tensor) -> Tensor:
...
+        stats_list: Optional[List[Dict[StatType, Any]]] = None,
+        stype: Optional[stype] = None,
+        post_module: Optional[Module] = None,
+        na_strategy: Optional[NAStrategy] = None,
+        in_channels: Optional[int] = None,
+        model: Optional[torch.nn.Module] = None,
+    ):
+        if in_channels is None:
+            raise ValueError("Please manually specify the `in_channels`, "
+                             "which is the text embedding dimensionality.")
```
vid-koci, line 630:
```suggestion
                             "specifying the text embedding dimensionality.")
```
Why do we make these arguments optional if they're not, in fact, optional? Can't we just drop these errors and make these args positional?

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -590,3 +594,71 @@ def encode_forward(self, feat: Tensor) -> Tensor:
...
+        super().init_modules()
+        num_cols = len(self.stats_list)
+        self.weight = Parameter(
+            torch.empty(num_cols, self.in_channels, self.out_channels))
+        self.bias = Parameter(torch.empty(num_cols, self.out_channels))
+        self.reset_parameters()
+
+    def reset_parameters(self):
+        super().reset_parameters()
+        torch.nn.init.normal_(self.weight, std=0.01)
```
vid-koci, line 652: Why this std? Why not use Xavier normal distribution that picks std on in/out dimension?

vid-koci (approved):
Left a few comments, otherwise, nice work!

torch_frame/data/tensor_frame.py
zechengz: Searched around, it seems that types from typing is mainly for hints. Usually in isinstance, concrete types such as `dict` is recommended to be used.

torch_frame/nn/encoder/stype_encoder.py
zechengz: If we want to make model not `optional`, we need to move it before arguments with default values. But it seems that then it will give some error due to `Module` initialization.

torch_frame/nn/encoder/stype_encoder.py
zechengz: To make them not optional we need to move it before the arguments with default values. But then it will be conflict with `Module` initialization and thus cause some error during `validate` cc @weihua916 We may need to improve this in the future.

torch_frame/nn/encoder/stype_encoder.py
zechengz: I just copied from what we previously have in other encoder. I think for PyTorch Linear the default one should be Kaiming uniform https://github.com/pytorch/pytorch/blob/498a76080299d44a3e22787bc42ec8e2eba71905/torch/nn/modules/linear.py#L107 @weihua916 any idea why we have this std with normal distribution?

torch_frame/data/mapper.py
zechengz: It depends on for example what tensor is returned by tokenizer. When we have multiple sentences, tokenizer will return a tensor with dim=2, each row corresponds to a row. Sometimes tokenizer might return 1 dim tensor when input (minibatch size is 1) only has one sentence.

test/nn/encoder/test_stype_encoder.py
```diff
@@ -246,3 +250,55 @@ def test_text_embedding_encoder():
...
+        model=model,
+    )
+    x_text = tensor_frame.feat_dict[stype.text_tokenized]
+    x = encoder(x_text)
+    assert x.shape == (num_rows,
+                       len(tensor_frame.col_names_dict[stype.text_tokenized]),
+                       out_channels)
+
+
+class SimpleTextModel(torch.nn.Module):
```
weihua916, line 294: `RandomTextModel`

examples/ft_transformer_text.py
```diff
@@ -51,21 +52,147 @@ def __call__(self, sentences: List[str]) -> Tensor:
...
+                inference_mode=False,
+                lora_dropout=0.1,
+                bias='none',
+                target_modules=target_modules,
+            )
+            self.model = get_peft_model(self.model, peft_config)
+        self.pooling = pooling
+
+    def forward(self, feat: TensorData) -> Tensor:
+        # [batch_size, num_cols, out_channels]
```
weihua916, line 135: What does this comment mean? shouldn't it be `[batch_size, num_cols, max_seq_len]`?

examples/ft_transformer_text.py
```diff
@@ -51,21 +52,147 @@ def __call__(self, sentences: List[str]) -> Tensor:
...
+        res = []
+        for s in sentences:
+            inputs = self.tokenizer(s, truncation=True, padding=False,
+                                    return_tensors='pt')
+            res.append(inputs)
+        return res
+
+
+class TextEncoder(torch.nn.Module):
+    def __init__(self, model: str, pooling: str = 'mean', lora: bool = False):
```
weihua916, line 109: Can we have some doc-string for `TextEncoder`? especially, what do you mean by `model`? What kind of string do you expect?

examples/ft_transformer_text.py
```diff
@@ -51,21 +52,147 @@ def __call__(self, sentences: List[str]) -> Tensor:
...
+class TextTokenizer:
+    def __init__(self, model: str):
+        self.tokenizer = AutoTokenizer.from_pretrained(model)
+
+    def __call__(self, sentences: List[str]) -> List[Dict[str, Tensor]]:
+        res = []
+        for s in sentences:
+            inputs = self.tokenizer(s, truncation=True, padding=False,
+                                    return_tensors='pt')
+            res.append(inputs)
```
weihua916, line 104: Here we are iterating over each sequence in the list. Can we use the batched tokenizer here?

examples/ft_transformer_text.py
```diff
@@ -51,21 +52,147 @@ def __call__(self, sentences: List[str]) -> Tensor:
...
+parser.add_argument('--pooling', type=str, default='mean',
+                    choices=['mean', 'cls'])
 args = parser.parse_args()
 
+if args.lora and not args.finetune:
+    raise ValueError('Please also specify finetune when '
+                     'choosing LoRA finetuning.')
+
+
+class PretrainedTextEncoder:
```
weihua916, line 71: `TextToEmbedding`

examples/ft_transformer_text.py
```diff
@@ -51,21 +52,147 @@ def __call__(self, sentences: List[str]) -> Tensor:
...
+    def __call__(self, sentences: List[str]) -> List[Dict[str, Tensor]]:
+        res = []
+        for s in sentences:
+            inputs = self.tokenizer(s, truncation=True, padding=False,
+                                    return_tensors='pt')
+            res.append(inputs)
+        return res
+
+
+class TextEncoder(torch.nn.Module):
```
weihua916, line 108: `TokenToEmbedding`, just to contrast with what we had before

examples/ft_transformer_text.py
```diff
@@ -51,21 +52,147 @@ def __call__(self, sentences: List[str]) -> Tensor:
...
+                lora_alpha=32,
+                inference_mode=False,
+                lora_dropout=0.1,
+                bias='none',
+                target_modules=target_modules,
+            )
+            self.model = get_peft_model(self.model, peft_config)
+        self.pooling = pooling
+
+    def forward(self, feat: TensorData) -> Tensor:
```
weihua916, line 134: I think this interface is nice! We should write a tutorial on this so that people understand how to write their model.

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -590,3 +594,71 @@ def encode_forward(self, feat: Tensor) -> Tensor:
         # -> [batch_size, num_cols, out_channels]
         x = x_lin + self.bias
         return x
+
+
+class LinearEmbeddingModelEncoder(StypeEncoder):
```
weihua916, line 599: How about `LinearModelEncoder` to make it shorter?

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -590,3 +594,71 @@ def encode_forward(self, feat: Tensor) -> Tensor:
...
+    r"""Linear function based encoder with a specified model outputs embedding
+    features. It applies a linear layer :obj:`torch.nn.Linear(in_channels,
+    out_channels)` on each embedding feature (:obj:`in_channels` is the
+    dimensionality of the embedding) and concatenates the output embeddings.
+    Note that the implementation does this for all :obj:`text_tokenized`
+    features in a batched manner.
```
weihua916, line 605: Can we mention that the model parameters are trained together?

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -590,3 +594,71 @@ def encode_forward(self, feat: Tensor) -> Tensor:
...
+        self.bias = Parameter(torch.empty(num_cols, self.out_channels))
+        self.reset_parameters()
+
+    def reset_parameters(self):
+        super().reset_parameters()
+        torch.nn.init.normal_(self.weight, std=0.01)
+        torch.nn.init.zeros_(self.bias)
+
+    def encode_forward(self, feat: TensorData) -> Tensor:
+        x = self.model(feat)
```
weihua916, line 656: I really like this interface! thanks. This makes it very easy to integrate other models on moldules/images etc.

examples/ft_transformer_text.py
weihua916: In https://huggingface.co/docs/transformers/main_classes/tokenizer, it says "a significant speed-up in particular when doing batched tokenization".

weihua916 (approved):
Great work, thanks! Left some comments.

I have two requests that I hope to be addressed after this PR.
- Support batched tokenization. 
- Properly handle NaN

torch_frame/nn/encoder/stype_encoder.py
weihua916: Also, can we specify what's the input and output of the `model`?

examples/ft_transformer_text.py
zechengz: Right, it's not `out_channels`, but should be `batch_max_seq_len` as we pad to max length for each batch.

examples/ft_transformer_text.py
zechengz: Maybe just `TextEmbedding`?

examples/ft_transformer_text.py
zechengz: Sure, I think we can update the docs later.

examples/ft_transformer_text.py
zechengz: Sure, will change to batch

test/datasets/test_fake_dataset.py
zechengz: Merge for now to unblock Weihua.

torch_frame/data/dataset.py
zechengz: Sure

PR #175, opened 2023-11-05
Handle the case where target contains nans in CatToNumTransform

### LINKED ISSUE #172: Handle the case when target_mean is nan in CatToNumTransform
Currently, it will just cast all the transformed values to nan

yiweny:
Fixes #172

### DIFF
--- a/test/transforms/cat_to_num_transform.py
+++ b/test/transforms/cat_to_num_transform.py
@@ -1,4 +1,3 @@
-import numpy as np
 import pytest
 import torch
 import torch.nn.functional as F
@@ -70,7 +69,7 @@ def test_cat_to_num_transform_on_categorical_only_dataset(with_nan):
 ])
 def test_cat_to_num_transform(task_type):
     num_rows = 10
-    dataset: Dataset = FakeDataset(num_rows=num_rows, with_nan=False,
+    dataset: Dataset = FakeDataset(num_rows=num_rows, with_nan=True,
                                    stypes=[stype.numerical, stype.categorical],
                                    task_type=task_type, create_split=True)
     dataset.df['x'] = 0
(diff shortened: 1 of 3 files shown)


torch_frame/transforms/cat_to_num_transform.py
```diff
@@ -59,6 +59,11 @@ def _fit(
         else:
             num_classes = 2
             target = tf_train.y.unsqueeze(1)
+            mask = ~torch.isnan(target)
+            if torch.isnan(target).any():
```
zechengz, line 63: nit
```suggestion
            if (~mask).sum() > 0:
```

zechengz (approved):

torch_frame/transforms/cat_to_num_transform.py
weihua916: (~mask).any()

test/transforms/cat_to_num_transform.py
```diff
@@ -123,3 +123,17 @@ def test_ordered_target_statistics_encoder(task_type):
...
             out.col_names_dict[stype.numerical][:total_numerical_cols])
+
+
+def test_cat_to_num_transform_with_target_contains_nans():
+    num_rows = 10
+    dataset: Dataset = FakeDataset(num_rows=num_rows, with_nan=False,
+                                   stypes=[stype.numerical, stype.categorical],
+                                   task_type=TaskType.REGRESSION,
+                                   create_split=True)
+    dataset.df.at[0, 'target'] = np.nan
```
weihua916, line 134: Why don't we modify `FakeDataset` so that `with_nan=True` gives nan values in the target as well>

yiweny: Makes sense. I hope making this change doesn't affect other test cases. 🥹

PR #177, opened 2023-11-05
add save and load logic to fittable transform

### LINKED ISSUE #77: Add saving/loading logic for fittable transform
We will need a saving/loading logic for fittable transform so that we don't fit every time. It's like Pytorch's model checkpoint saving/loading.

yiweny:
Fixes #77


torch_frame/transforms/fittable_base_transform.py
```diff
@@ -90,3 +98,11 @@ def _fit(self, tf: TensorFrame, col_stats: Dict[str, Dict[StatType, Any]]):
     @abstractmethod
     def _forward(self, tf: TensorFrame) -> TensorFrame:
         raise NotImplementedError
+
+    @abstractmethod
+    def _save(self, path: str) -> None:
+        raise NotImplementedError
+
+    @abstractmethod
+    def _load(self) -> None:
```
rusty1s, line 107: This should be a `classmethod`.

torch_frame/transforms/fittable_base_transform.py
```diff
@@ -90,3 +98,11 @@ def _fit(self, tf: TensorFrame, col_stats: Dict[str, Dict[StatType, Any]]):
     @abstractmethod
     def _forward(self, tf: TensorFrame) -> TensorFrame:
         raise NotImplementedError
+
+    @abstractmethod
+    def _save(self, path: str) -> None:
```
rusty1s, line 103: Why does this need an underscore?

weihua916 (commented):
Thanks for starting this!

Ideally, what we want is the following:

```python
transform = CatToNumTransform()
transform.fit(dataset.tensor_frame, dataset.col_stats)
# save the checkpoint
torch.save(transform.state_dict(), path)

# Initialize a completely new transform
loaded_transform = CatToNumTransform()
# Load from checkpoint
loaded_transform.load_state_dict(torch.load(path))
# Make sure loaded_transform and transform behaves the same
```

weihua916 (approved):
Looks good. Made some edits to simplify the test.

Issue #88, opened 2023-09-29
More stype support


weihua916:
We aim to expand our scope of Pytorch Frame beyond the basic `numerical` and `categorical` columns.

`stype.text_embedded` @zechengz @weihua916 
- [x] `TensorMapper`
- [x] `StypeEncoder`
- [x] e2e example

`stype.text_tokenized` @zechengz
- [x] `TensorMapper`
- [x] `StypeEncoder`
- [x] e2e example

`stype.timestamp` @yiweny  
- [x] `TensorMapper`
- [x] `StypeEncoder`
- [x] e2e example

`stype.sequence_numerical` @yiweny 
- [x] `TensorMapper`
- [ ] `StypeEncoder`
- [ ] e2e example

`stype.sequence_categorical` @yiweny 
- [ ] `TensorMapper`
- [ ] `StypeEncoder`
- [ ] e2e example

`stype.multi_categorical` @yiweny 
- [x] `TensorMapper`
- [x] `StypeEncoder`
- [x] e2e example

`stype.embeddings` @akihironitta 
- [x] `MultiEmbeddingTensor`
- [x] `TensorMapper`
- [x] `StypeEncoder`
- [x] e2e example

PR #181, opened 2023-11-06
[2/n] `MultiEmbeddingTensor`: Add `MultiEmbeddingTensor`

akihironitta:
Adds `MultiEmbeddingTensor` as part of #88.

Follow-up tasks:
- Support slicing and indexing
- Add `stype.embedding` and `EmbeddingTensorMapper`
- Add an example
- ~Refactor this with `MultiNestedTensor`~ being addressed via #183.


torch_frame/data/multi_embedding_tensor.py
```diff
@@ -0,0 +1,166 @@
+import copy
+from typing import Any, Callable, Sequence, Tuple
+
+import torch
+from torch import Tensor
+
+
+class MultiEmbeddingTensor:
+    r"""A PyTorch tensor-based data structure that stores
+    :obj:`[num_rows, num_cols, *]`, where the size of last dimension can be
```
zechengz, line 10: QQ: the dimension for the same column should be same right?

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -0,0 +1,166 @@
...
+                        "num_cols must be the same across a list of input "
+                        "multi nested tensors.")
+            values = torch.cat([x.values for x in xs], dim=0)
+            return MultiEmbeddingTensor(
+                num_rows=num_rows,
+                num_cols=num_cols,
+                values=values,
+                offset=xs[0].offset,
+            )
+        else:
```
zechengz, line 150: `elif dim == 1`? Also should we support cat to dim == 2?

torch_frame/data/multi_embedding_tensor.py
rusty1s: Yes. I think it would be great to highlight in the doc what's the difference between `MultiNestedTensor` and `MultiEmbeddingTensor`.

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -0,0 +1,166 @@
...
+        assert offset[0] == 0
+        assert len(offset) == num_cols + 1
+        self.num_rows = num_rows
+        self.num_cols = num_cols
+        self.values = values
+        self.offset = offset
+
+    def __setitem__(self, index: Any, values: Any) -> None:
+        raise RuntimeError(
+            "MultiEmbeddingTensor object does not support setting values. "
```
rusty1s, line 37:
```suggestion
            "{self.__class__.__name__} object does not support setting values. "
```

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -0,0 +1,166 @@
...
+
+    Args:
+        num_rows (int): Number of rows.
+        num_cols (int): Number of columns.
+        values (torch.Tensor): The values :class:`torch.Tensor` of size
+            :obj:`[num_rows, dim1+dim2+...+dimN]`.
+        offset (torch.Tensor): The offset :class:`torch.Tensor` of size
+            :obj:`[num_cols+1,]`.
+    """
+    def __init__(
```
rusty1s, line 21: I feel a lot of stuff here can be re-used from `MultiNestedTensor`. Does it make sense to create a base class here?

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -0,0 +1,166 @@
...
+        self,
+        num_rows: int,
+        num_cols: int,
+        values: Tensor,
+        offset: Tensor,
+    ) -> None:
```
rusty1s, line 27: Don't we need a `classmethod` here to construct `MultiEmbeddingTensor` much more easily, e.g.:
```python
MultiEmbeddingTensor.from_list(embeddings: List[Tensor])` where `len(embeddings) == num_cols`.

File: torch_frame/data/multi_embedding_tensor.py [hunk 6]
```diff
@@ -0,0 +1,166 @@
...
+        else:
+            num_rows = xs[0].num_rows
+            num_cols = sum([x.num_cols for x in xs])
+            offset = torch.empty(
+                num_cols + 1,
+                dtype=torch.long,
+                device=device,
+            )
+            offset[0] = 0
+            offset[1:] = torch.cat([x.offset[1:] for x in xs], dim=0)
```
rusty1s, line 159: Where do we accumulate offsets?

test/data/test_multi_embedding_tensor.py
```diff
@@ -0,0 +1,75 @@
+import pytest
+import torch
+
+from torch_frame.data.multi_embedding_tensor import MultiEmbeddingTensor
+
+
+def test_size():
```
weihua916, line 7: Looks like you are starting from `values` and `offset`, which is internal storage of multiple embedding tensors.

For testing, can we follow the pattern [here](https://github.com/pyg-team/pytorch-frame/blob/master/test/data/test_multi_nested_tensor.py#L35-L44), where you implement
```python
# Create `tensor_mat: List[Tensor]`. A list of tensors of shape [num_rows, emb_dim] for different columns.
met = MultiEmbeddingTensor.from_tensor_mat(tensor_mat)
```

torch_frame/data/multi_embedding_tensor.py
weihua916: +1. I think we can follow the pattern of `MultiNestedTensor`

torch_frame/data/multi_embedding_tensor.py
weihua916: Yes, I think apply(), device() logics can definitely be shared. getitem, index logics will be specially implemented for this `MultiNestedTensor`

weihua916 (commented):
Thanks for starting. I think we can create a base class to share basic logics with `MultiNestedTensor`. For `MultiEmbeddingTensor`, I think you can first implement `MultiEmbeddingTensor.from_tensor_mat` and make sure `met[i, j]` is working properly. Then, we can gradually build up advanced indexing and concatenation in separate PRs.

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -0,0 +1,166 @@
...
+
+    @classmethod
+    def _check_dim(cls, dim: int) -> int:
+        r"""Check :obj:`dim` argument and make it 0 or 1."""
+        if dim < 0:
+            dim = dim + 3
+        if dim not in [0, 1]:
+            raise ValueError(
+                f"Advanced indexing with dim={dim} is unsupported in "
+                f"MultiEmbeddingTensor. Please use dim=0 or 1.")
```
yiweny, line 48: raise a different error when dim > 2

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -0,0 +1,166 @@
...
+        return 3
+
+    def dim(self) -> int:
+        return self.ndim
+
+    def size(self, dim: int) -> int:
+        r"""Dimension of the :class:`torch_frame.data.MultiEmbeddingTensor`"""
+        # Handle negative dim
+        if dim < 0:
+            dim = self.ndim - dim
```
yiweny, line 101:
```suggestion
            dim = self.ndim + dim
```

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -0,0 +1,166 @@
...
+    def __setitem__(self, index: Any, values: Any) -> None:
+        raise RuntimeError(
+            "MultiEmbeddingTensor object does not support setting values. "
+            "It should be used for read-only.")
+
+    @classmethod
+    def _check_dim(cls, dim: int) -> int:
+        r"""Check :obj:`dim` argument and make it 0 or 1."""
+        if dim < 0:
+            dim = dim + 3
```
yiweny, line 44:
```suggestion
            dim = dim + self.ndim
```
Please don't use magic numbers.

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -0,0 +1,166 @@
...
+        return self.values.dtype
+
+    # Static methods ##########################################################
+    @staticmethod
+    def cat(
+        xs: Sequence['MultiEmbeddingTensor'],
+        dim: int = 0,
+    ) -> 'MultiEmbeddingTensor':
+        if len(xs) == 0:
+            raise RuntimeError("Cannot concatenate a list of length 0.")
```
yiweny, line 131: `Sequence` is not limited to list.

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -0,0 +1,166 @@
...
+        assert isinstance(xs[0], MultiEmbeddingTensor)
```
yiweny, line 132: Check this for all items.

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -0,0 +1,166 @@
...
+        dim = MultiEmbeddingTensor._check_dim(dim)
+        device = xs[0].device
```
yiweny, line 134: You need to check this for all items in xs. There's no guarantee that they are all going to be on the same device.

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -0,0 +1,166 @@
...
+        if dim == 0:
+            num_rows = sum(x.num_rows for x in xs)
+            num_cols = xs[0].num_cols
+            for x in xs[1:]:
+                if x.num_cols != num_cols:
+                    raise RuntimeError(
+                        "num_cols must be the same across a list of input "
+                        "multi nested tensors.")
+            values = torch.cat([x.values for x in xs], dim=0)
```
yiweny, line 143: You probably want to check if all items in xs are in the same dtype. If not, probably good to raise an error or a warning.

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -0,0 +1,166 @@
...
+        dim: int = 0,
+    ) -> 'MultiEmbeddingTensor':
+        if len(xs) == 0:
+            raise RuntimeError("Cannot concatenate a list of length 0.")
+        assert isinstance(xs[0], MultiEmbeddingTensor)
+        dim = MultiEmbeddingTensor._check_dim(dim)
+        device = xs[0].device
+        if dim == 0:
+            num_rows = sum(x.num_rows for x in xs)
+            num_cols = xs[0].num_cols
```
yiweny, line 137: Validate if all items in xs has the same number of columns.

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -0,0 +1,166 @@
...
+                        "multi nested tensors.")
+            values = torch.cat([x.values for x in xs], dim=0)
+            return MultiEmbeddingTensor(
+                num_rows=num_rows,
+                num_cols=num_cols,
+                values=values,
+                offset=xs[0].offset,
+            )
+        else:
+            num_rows = xs[0].num_rows
```
yiweny, line 151: Validate if all items in xs has the same rows.

torch_frame/data/multi_embedding_tensor.py
yiweny: + 1

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -0,0 +1,166 @@
...
+            num_cols = sum([x.num_cols for x in xs])
+            offset = torch.empty(
+                num_cols + 1,
+                dtype=torch.long,
+                device=device,
+            )
+            offset[0] = 0
+            offset[1:] = torch.cat([x.offset[1:] for x in xs], dim=0)
+            values = torch.cat([x.values for x in xs], dim=1)
+            return MultiEmbeddingTensor(
```
yiweny, line 161: Curious to see if it's possible to do some garbage collection here? Or modify the existing `MultiEmbeddingTensor` in place? If the tensors are on gpu, it seems to me that they will remain on the gpu.

test/data/test_multi_embedding_tensor.py
```diff
@@ -0,0 +1,75 @@
...
+    assert met.shape[0] == num_rows
+    assert met.shape[1] == num_cols
+    assert met.shape[2] == -1
+
+
+def test_cat():
+    num_rows = 3
+    num_cols = 2
+    offset = torch.tensor([0, 2, 4])
+    values = torch.tensor([
```
yiweny, line 36: Use `rand` to generate tensor.

test/data/test_multi_embedding_tensor.py
```diff
@@ -0,0 +1,75 @@
...
+        [0, 1, 2, 3, 4],
+        [5, 6, 7, 8, 9],
+        [10, 11, 12, 13, 14],
+    ], )
+    assert torch.allclose(met_cat.values, expected_values)
+    expected_offset = offset
+    assert torch.allclose(met_cat.offset, expected_offset)
+
+    # Test column concat
+    met_cat = MultiEmbeddingTensor.cat([met, met], dim=1)
```
yiweny, line 65: Test the case with `(met, met)`

test/data/test_multi_embedding_tensor.py
```diff
@@ -0,0 +1,75 @@
...
+    assert met_cat.num_rows == num_rows
+    assert met_cat.num_cols == num_cols * 2
+    expected_values = torch.tensor([
+        [0, 1, 2, 3, 4, 0, 1, 2, 3, 4],
+        [5, 6, 7, 8, 9, 5, 6, 7, 8, 9],
+        [10, 11, 12, 13, 14, 10, 11, 12, 13, 14],
+    ], )
+    assert torch.allclose(met_cat.values, expected_values)
+    expected_offset = torch.tensor([0, 2, 4, 2, 4])
+    assert torch.allclose(met_cat.offset, expected_offset)
```
yiweny, line 75: Add more test cases to edge case, e.g. the tensors you'd like to concat are different shapes, different types, different device.

yiweny (changes requested):
Thank you for your work. I think there are some small bugs and edge case handling that need to be resolved in this pr.
Also, I think it might be better to create a base class for this and `MultiNestedTensor`.

test/data/test_multi_embedding_tensor.py
akihironitta: Resolving this comment as I decided to add `cat` in a follow-up PR.

test/data/test_multi_embedding_tensor.py
akihironitta: Resolving this comment as I decided to add `cat` in a follow-up PR.

torch_frame/data/multi_embedding_tensor.py
akihironitta: Users will still have the reference to the tensors, so I think there's nothing we should do on our side here.

torch_frame/data/multi_embedding_tensor.py
akihironitta: Already done below this line.

torch_frame/data/multi_embedding_tensor.py
akihironitta: Users may find `cat(..., dim==2)` useful. Will address in a follow-up :)

torch_frame/data/multi_embedding_tensor.py
akihironitta: Addressing via #184

torch_frame/data/multi_embedding_tensor.py
akihironitta: Briefly described the difference between these two. PTAL :)

torch_frame/data/multi_embedding_tensor.py
akihironitta: DOne!

torch_frame/data/multi_embedding_tensor.py
akihironitta: Following tensor operations will error out anyway, but added an assertion on our side as you suggest.

torch_frame/data/multi_embedding_tensor.py
akihironitta: Addressed your comment in #183

test/data/test_multi_embedding_tensor.py
akihironitta: I prefer this way to random tensors for readability, but I can replace this if you'd advocate for it

akihironitta (commented):
@yiweny @rusty1s @weihua916 @zechengz Ready for your second review ;)

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -0,0 +1,116 @@
...
+        offset_list = []
+        accum_idx = 0
+        offset_list.append(accum_idx)
+        for tensor in tensor_list:
+            accum_idx += tensor.size(1)
+            offset_list.append(accum_idx)
+
+        num_cols = len(tensor_list)
+        values = torch.cat(tensor_list, dim=1)
+        assert values.size() == (num_rows, offset_list[-1])
```
weihua916, line 114: Shall we move this assert inside init?

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -0,0 +1,116 @@
...
+                index[0], int) and isinstance(index[1], int):
+            i = index[0]
+            j = index[1]
+            return self.values[i, self.offset[j]:self.offset[j + 1]]
+
+        # TODO(akihironitta): Support more index types
+        raise NotImplementedError
+
+    @classmethod
+    def from_list(
```
weihua916, line 64: Can we rename this to `from_tensor_list`? This aligns better with `from_tensor_mat` in `MultiNestedTensor`.

weihua916 (approved):
LGM, thanks!

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -0,0 +1,116 @@
...
+        offset: Tensor,
+    ) -> None:
+        super().__init__(num_rows, num_cols, values, offset)
+        assert offset[0] == 0
+        assert len(offset) == num_cols + 1
+        assert values.size() == (num_rows, offset[-1])
+
+    def __getitem__(
+        self,
+        index: Any,
```
yiweny, line 53:
```suggestion
        index: IndexSelectType,
```

yiweny (approved):

torch_frame/data/multi_embedding_tensor.py
akihironitta: Will address in a follow-up!

PR #182, opened 2023-11-06
Add `Amazon Fine Food Reviews` and OpenAI embedding example

zechengz:
Add `Amazon Fine Food Reviews` dataset and OpenAI embedding example as it's also an example in OpenAI https://platform.openai.com/docs/guides/embeddings/use-cases


examples/openai_embedding.py
```diff
@@ -0,0 +1,152 @@
...
+    LinearEmbeddingEncoder,
+    LinearEncoder,
+)
+
+# Please run `export OPENAI_API_KEY=...` before running the code
+# Notice that there are 568,454 rows and 2 text columns, it will
+# cost some money to get the text embeddings by using OpenAI API
+
+
+class OpenAIEmbedding:
```
rusty1s, line 28: Does there exist a base class for this?

examples/openai_embedding.py
```diff
@@ -0,0 +1,152 @@
...
+device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
+
+# Prepare datasets
+path = osp.join(osp.dirname(osp.realpath(__file__)), '..', 'data',
+                'amazon_fine_food_reviews')
+text_encoder = OpenAIEmbedding(model=args.model)
+dataset = AmazonFineFoodReviews(
+    root=path,
+    text_embedder_cfg=TextEmbedderConfig(text_embedder=text_encoder,
+                                         batch_size=5),
```
rusty1s, line 63:
```suggestion
                                         batch_size=5, ),
```

examples/openai_embedding.py
```diff
@@ -0,0 +1,152 @@
...
+
+# Shuffle the dataset
+dataset = dataset.shuffle()
+train_dataset, val_dataset, test_dataset = dataset[:0.8], dataset[
+    0.8:0.9], dataset[0.9:]
+
+# Set up data loaders
+train_tensor_frame = train_dataset.tensor_frame
+val_tensor_frame = val_dataset.tensor_frame
+test_tensor_frame = test_dataset.tensor_frame
```
rusty1s, line 76: We can just input the dataset into `DataLoader`.

examples/openai_embedding.py
```diff
@@ -0,0 +1,152 @@
...
+    stype.categorical: EmbeddingEncoder(),
+    stype.numerical: LinearEncoder(),
+    stype.text_embedded: LinearEmbeddingEncoder(in_channels=args.emb_size)
+}
+
+output_channels = dataset.num_classes
+
+model = FTTransformer(
+    channels=args.channels,
+    out_channels=output_channels,
```
rusty1s, line 92:
```suggestion
model = FTTransformer(
    channels=args.channels,
    out_channels=dataset.num_classes,
```

examples/openai_embedding.py
```diff
@@ -0,0 +1,152 @@
...
+from torch_frame.nn import (
+    EmbeddingEncoder,
+    FTTransformer,
+    LinearEmbeddingEncoder,
+    LinearEncoder,
+)
+
+# Please run `export OPENAI_API_KEY=...` before running the code
+# Notice that there are 568,454 rows and 2 text columns, it will
+# cost some money to get the text embeddings by using OpenAI API
```
yiweny, line 25: Is it possible to supply it as an argument of the script. You can first check if it's supplied, then if it's in env var. If not, you error out.

yiweny (approved):
LGTM. Thanks

weihua916 (approved):
Thanks. Curious how well it works

examples/openai_embedding.py
zechengz: For now I think no?

examples/openai_embedding.py
zechengz: Sure

PR #183, opened 2023-11-06
[1/n] `MultiEmbeddingTensor`: Restructure and clean up `MultiNestedTensor`

akihironitta:
Preparation for #181.

- Adds a base class for `Multi{Nested,Embedding}Tensor`.
- Fixes typing.
- Adds more docs.
- Make a few methods private.


torch_frame/data/multi_nested_tensor.py
```diff
@@ -116,41 +116,34 @@ def from_tensor_mat(
...
                 # Returns Tensor
-                return self.get_value(index[0], index[1])
+                return self._get_value(index[0], index[1])
             else:
                 # Returns MultiNestedTensor
-                out = self
                 for dim, idx in enumerate(index):
-                    out = out.select(idx, dim)
-                return out
+                    self = self.select(idx, dim)
```
zechengz, line 132: Does this mean that we will modify `self` when doing some indexing?

yiweny: +1

torch_frame/data/multi_tensor.py
```diff
@@ -0,0 +1,107 @@
...
+
+    @classmethod
+    def _normalize_dim(cls, dim: int) -> int:
+        r"""Normalize :obj:`dim` argument from range [-2, 1] to range [0, 1].
+
+        Raises:
+            IndexError: If :obj:`dim` is out of range [-2, 1].
+        """
+        if dim < 0:
+            dim = dim + 3
```
yiweny, line 97: Remove magic number

torch_frame/data/multi_tensor.py
```diff
@@ -0,0 +1,107 @@
...
+        if dim == 2:
+            raise IndexError(
+                "Dimension out of range (expected to be in range of [-2, 1],"
```
yiweny, line 100: Prolly not need to add the first line to the error

yiweny (approved):

torch_frame/data/multi_nested_tensor.py
weihua916: It seems that we are not modifying self in-place, since the unit test is passing. @akihironitta could you double-check?

weihua916 (approved):
Very clean. Thanks!

akihironitta:
I'll address any other review in a follow-up #181 if there is.

PR #186, opened 2023-11-08
add cohere embedding example

yiweny:


examples/ft_transformer_text.py
```diff
@@ -1,7 +1,9 @@
 import argparse
+import os
 import os.path as osp
 from typing import List
 
+import cohere
```
zechengz, line 6: Move this to the cohere embedding class? Now it requires user install cohere to run the text example.

zechengz (commented):
Left some comment

weihua916:
Thanks! How about we merge this with https://github.com/pyg-team/pytorch-frame/blob/master/examples/openai_embedding.py, since both are using API?
and rename to `text_api_embeddings.py`?

examples/ft_transformer_text.py
yiweny: Makes sense

yiweny:
> Thanks! How about we merge this with https://github.com/pyg-team/pytorch-frame/blob/master/examples/openai_embedding.py, since both are using API? and rename to `text_api_embeddings.py`?

But I want to put the experiment results we produced yesterday somewhere. Or we can put it in the readme?

weihua916:
Got it, but I think it's cleaner to have well-organized example code according to their functionality. I don't think where you keep results is so important.

yiweny:
> Got it, but I think it's cleaner to have well-organized example code according to their functionality. I don't think where you keep results is so important.

Resolved~

examples/openai_embedding.py
```diff
@@ -28,21 +26,33 @@
 parser.add_argument('--lr', type=float, default=0.0001)
 parser.add_argument('--epochs', type=int, default=100)
 parser.add_argument('--seed', type=int, default=0)
+parser.add_argument('--service', type=str, default='cohere',
+                    choices=['openai', 'cohere'])
 parser.add_argument('--model', type=str, default='text-embedding-ada-002')
-parser.add_argument('--emb_size', type=int, default=1536)
 parser.add_argument('--openai_api_key', type=str, default=None)
+parser.add_argument('--cohere_api_key', type=str, default=None)
```
weihua916, line 33: nit: why not just have `api_key`?

examples/openai_embedding.py
```diff
@@ -28,21 +26,33 @@
 parser.add_argument('--lr', type=float, default=0.0001)
 parser.add_argument('--epochs', type=int, default=100)
 parser.add_argument('--seed', type=int, default=0)
+parser.add_argument('--service', type=str, default='cohere',
+                    choices=['openai', 'cohere'])
 parser.add_argument('--model', type=str, default='text-embedding-ada-002')
```
weihua916, line 31: Can we drop model argument since it's special to openai?

weihua916 (approved):
Looks great. Logo as well :)

README.md
```diff
@@ -183,6 +187,34 @@ for epoch in range(50):
...
+      <a href="https://platform.openai.com/docs/guides/embeddings">
+        <img src="docs/source/_figures/OpenAI_Logo.png" alt="OpenAI" width="100px"/>
+      </a>
+      <br /><a href="https://github.com/pyg-team/pytorch-frame/blob/master/examples/openai_embedding.py">OpenAI Embedding Code Example</a>
+    </td>
+    <td align="center">
+      <a href="https://cohere.com/embeddings">
+        <img src="docs/source/_figures/cohere-logo.png" alt="Cohere" width="100px"/>
+      </a>
+      <br /><a href="https://github.com/pyg-team/pytorch-frame/blob/master/examples/openai_embedding.py">Cohere Embed v3 Code Example</a>
```
zechengz, line 207: Rename the `openai_embedding.py` to `text_api_embeddings.py` as mentioned by Weihua and change both openai and cohere example path to that?

yiweny: I want to call it `llm_embeddings.py` if that's fine with you

zechengz: I am ok. Though it sounds too fancy to me 😂

PR #190, opened 2023-11-09
Support `TensorFrame` `concat` for complex stypes + test clean-up

weihua916:
Also, add more thorough testing of all complex stypes that we've added recently.


torch_frame/data/tensor_frame.py
```diff
@@ -168,26 +169,28 @@ def __eq__(self, other: Any) -> bool:
...
-                    if isinstance(self_feat_value,
-                                  Tensor) and not torch.allclose(
-                                      self_feat_value, other_feat_value):
+            elif isinstance(self_feat, MultiNestedTensor):
+                if not isinstance(other_feat, MultiNestedTensor):
+                    return False
+                if not MultiNestedTensor.allclose(self_feat, other_feat):
+                    return False
+            elif isinstance(self_feat, Dict):
+                if not isinstance(other_feat, Dict):
```
akihironitta, line 187: QQ: Any idea what the difference from built-in `dict` is when evaluating object with `isinstance`?
```suggestion
            elif isinstance(self_feat, dict):
                if not isinstance(other_feat, dict):
```

akihironitta (approved):
Neat!

torch_frame/data/tensor_frame.py
weihua916: I think they have the same behavior.
```python
>>> from typing import Dict
>>> isinstance({}, Dict)
True
>>> isinstance({}, dict)
True
>>>
```

akihironitta: (cc @yiweny as I remember you also left a similar comment on my previous PR: `isinstance(obj, Tuple)` vs. `isinstance(obj, tuple)`.)

PR #193, opened 2023-11-09
[3/n] `MultiEmbeddingTensor`: Support `cat`

akihironitta:
Part of #88.


test/data/test_multi_embedding_tensor.py
```diff
@@ -117,3 +118,130 @@ def test_clone():
...
+        torch.tensor([[5], [11]]),
+    ]
+    met1 = MultiEmbeddingTensor.from_tensor_list(tensor_list1)
+    tensor_list2 = [
+        torch.tensor([[12, 13, 14]]),
+        torch.tensor([[15, 16]]),
+        torch.tensor([[17]]),
+    ]
+    met2 = MultiEmbeddingTensor.from_tensor_list(tensor_list2)
+    for xs in [[met1, met2], (met1, met2)]:
```
akihironitta, line 137: Testing both `list`/`tuple` as requested in https://github.com/pyg-team/pytorch-frame/pull/181#discussion_r1383949747.

test/data/test_multi_embedding_tensor.py
```diff
@@ -117,3 +118,138 @@ def test_clone():
...
+        assert met_cat.num_cols == met1.num_cols + met2.num_cols
+        expected_values = torch.tensor([
+            [0, 1, 2, 3, 4, 5, 6, 7, 8, 9],
+            [10, 11, 12, 13, 14, 15, 16, 17, 18, 19],
+        ])
+        assert torch.allclose(met_cat.values, expected_values)
+        expected_offset = torch.tensor([0, 3, 5, 6, 10])
+        assert torch.allclose(met_cat.offset, expected_offset)
+
+    # case: dim=1 with different num_cols should raise error
```
akihironitta, line 200:
```suggestion
    # case: dim=1 with different num_rows should raise error
```

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -114,3 +114,87 @@ def from_tensor_list(
...
+            msg = "xs must be a list of MultiEmbeddingTensor."
+            assert isinstance(x, MultiEmbeddingTensor), msg
+            msg = ("device must be the same across a sequence of"
+                   " MultiEmbeddingTensor.")
+            assert x.device == xs[0].device, msg
+
+        dim = MultiEmbeddingTensor._normalize_dim(dim)
+
+        if len(xs) == 1:
+            return xs[0].clone()
```
weihua916, line 173: Do we need clone here, given that this is read-only?

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -114,3 +114,87 @@ def from_tensor_list(
...
+        if dim == 0:
+            num_rows = sum(x.num_rows for x in xs)
+            num_cols = xs[0].num_cols
+            for x in xs[1:]:
+                if x.num_cols != num_cols:
+                    raise RuntimeError(
+                        "num_cols must be the same across a list of input "
+                        "multi embedding tensors.")
+            values = torch.cat([x.values for x in xs], dim=0)
+            offset = xs[0].offset.clone()
```
weihua916, line 184: same. Do we need clone?

test/data/test_multi_embedding_tensor.py
```diff
@@ -117,3 +118,138 @@ def test_clone():
     assert met_clone.values[0, 0] != 12345.
     met.offset[0] = -1
     assert met_clone.offset[0] != -1
+
+
+def test_cat():
```
weihua916, line 123: I think this `cat`'s test is good, but a more direct way is some test like [this](https://github.com/pyg-team/pytorch-frame/blob/master/test/data/test_multi_nested_tensor.py#L164-L178). This requires slicing to be first implemented. Not a blocker, but I'd prefer swapping this `test_cat` with something like [this](https://github.com/pyg-team/pytorch-frame/blob/master/test/data/test_multi_nested_tensor.py#L164-L178) in the subsequent PRs (once slicing is done and tested.)

weihua916 (approved):
Thanks. Left a comment regarding the necessity of cloning and `cat`'s unit test.

torch_frame/data/multi_embedding_tensor.py
akihironitta: I cloned this because `torch.cat([tensor])` also behaves this way, but I'm open to reverting this.

test/data/test_multi_embedding_tensor.py
weihua916: My main concern with this testing approach is that the comparison in offset and values space is quite unintuitive.

torch_frame/data/multi_embedding_tensor.py
akihironitta: ditto

test/data/test_multi_embedding_tensor.py
akihironitta: I see, that totally makes sense :) I'll make these test cases consistent with the `MultiNestedTensor` ones in a follow-up PR adding slicing!

torch_frame/data/multi_embedding_tensor.py
weihua916: Got it, I'm open as well. How about removing clone for now, leaving comments that we don't clone since the tensor is read-only?

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -114,3 +114,87 @@ def from_tensor_list(
...
+            >>> out.values
+            tensor([[ 0,  1,  2,  3,  4,  5],
+                    [ 6,  7,  8,  9, 10, 11],
+                    [12, 13, 14, 15, 16, 17]])
+        """
+        if len(xs) == 0:
+            raise RuntimeError('Cannot concatenate a sequence of length 0.')
+
+        for x in xs:
+            msg = "xs must be a list of MultiEmbeddingTensor."
```
zechengz, line 164: nit:
```suggestion
            msg = "`xs` must be a list of MultiEmbeddingTensor."
```

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -114,3 +114,87 @@ def from_tensor_list(
...
+            assert isinstance(x, MultiEmbeddingTensor), msg
+            msg = ("device must be the same across a sequence of"
+                   " MultiEmbeddingTensor.")
+            assert x.device == xs[0].device, msg
```
zechengz, line 168: Maybe get list of all devices and check whether all of them are equal?

zechengz (approved):

test/data/test_multi_embedding_tensor.py
weihua916: Sounds good, thanks! the doc-string of `cat` can be also modified. We try not to expose `values` and `offset` since they are for internal storage.

torch_frame/data/multi_embedding_tensor.py
akihironitta: Removed the `clone()` call as per https://github.com/pyg-team/pytorch-frame/pull/193#discussion_r1387657414.

PR #194, opened 2023-11-09
Support `stype.embedding`

akihironitta:
Part of #88.


akihironitta:
This PR is on-hold until advanced indexing is supported in `MultiEmbeddingTensor`.

torch_frame/data/mapper.py
```diff
@@ -395,3 +396,31 @@ def forward(
...
+    def forward(
+        self,
+        ser: Series,
+        *,
+        device: Optional[torch.device] = None,
+    ) -> MultiEmbeddingTensor:
+        # NOTE We are converting the default PyTorch dtype into a string
+        # representation that can be understood by numpy.
+        # TODO Think of a "less hacky" way to do this.
+        dtype = str(torch.get_default_dtype()).split('.')[-1]
```
weihua916, line 412: For embedding, the dtype should always be float32?

torch_frame/data/mapper.py
```diff
@@ -395,3 +396,31 @@ def forward(
...
+        values = ser.explode().astype(dtype).values
+        values = torch.from_numpy(values).view(len(ser), -1)
```
weihua916, line 414:
```python
values = torch.from_numpy(np.stack(df.values))
```

torch_frame/data/stats.py
```diff
@@ -39,7 +39,7 @@ class StatType(Enum):
 
     # text_embedded (Also, embedding)
     # Note: For text_embedded, this stats is computed in
-    # dataset.update_col_stats, not here.
+    # dataset._update_col_stats, not here.
```
weihua916, line 42: We can follow-up in the next PR, but we need to compute `EMB_DIM` as `compute_stats` here for `stype.embedding`.

weihua916 (approved):
Looks great, thanks!

Left a comment on follow-up PR for `EMB_DIM` stats computation.

torch_frame/data/mapper.py
zechengz: Maybe we should keep this if user want to train the model with float16?

torch_frame/data/tensor_frame.py
```diff
@@ -186,6 +187,12 @@ def __eq__(self, other: Any) -> bool:
                 if not MultiNestedTensor.allclose(self_feat, other_feat,
                                                   equal_nan=True):
                     return False
+            elif isinstance(self_feat, MultiEmbeddingTensor):
+                if not isinstance(other_feat, MultiEmbeddingTensor):
+                    return False
+                if not MultiNestedTensor.allclose(self_feat, other_feat,
```
zechengz, line 193: Maybe this?
```suggestion
                if not MultiEmbeddingTensor.allclose(self_feat, other_feat,
```

torch_frame/typing.py
```diff
@@ -66,4 +67,4 @@ def is_numerical_strategy(self):
 TextTokenizationMapping = Mapping[str, Tensor]
 TextTokenizationOutputs = Union[List[TextTokenizationMapping],
                                 TextTokenizationMapping]
-TensorData = Union[Tensor, MultiNestedTensor, Dict[str, MultiNestedTensor]]
+TensorData = Union[Tensor, _MultiTensor, Dict[str, MultiNestedTensor]]
```
zechengz, line 70: Maybe union of MultiNestedTensor and MultiEmbeddingTensor is better?

zechengz (approved):

Issue #195, opened 2023-11-09
sequence numerical stats computation 


weihua916:
Currently, sequence numerical stats will be NaN as long as we have a single sequence with NaN element. This is because `dropna()` in https://github.com/pyg-team/pytorch-frame/blob/master/torch_frame/data/stats.py#L101 cannot drop NaN elements inside the sequence. 

We want to handle NaN in sequence gracefully and augment the test case [here](https://github.com/pyg-team/pytorch-frame/blob/master/test/data/test_stats.py#L39-L46).

PR #196, opened 2023-11-09
Fix stats computation for `sequence_numerical` with `NaN` elements

weihua916:
Fixes https://github.com/pyg-team/pytorch-frame/issues/195

### DIFF
--- a/torch_frame/data/stats.py
+++ b/torch_frame/data/stats.py
@@ -49,15 +49,27 @@ class StatType(Enum):
     def compute(self, ser: Series, sep: Optional[str] = None) -> Any:
         if self == StatType.MEAN:
             flattened = np.hstack(np.hstack(ser.values))
-            return np.mean(flattened).item()
+            finite_mask = np.isfinite(flattened)
+            if not finite_mask.any():
+                return np.nan
+            return np.mean(flattened[finite_mask])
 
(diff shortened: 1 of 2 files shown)


torch_frame/data/stats.py
```diff
@@ -49,15 +49,27 @@ def stats_for_stype(stype: torch_frame.stype) -> List['StatType']:
...
+            if not finite_mask.any():
+                return np.nan
+            return np.mean(flattened[finite_mask]).item()
 
         elif self == StatType.STD:
             flattened = np.hstack(np.hstack(ser.values))
-            return np.std(flattened).item()
+            finite_mask = np.isfinite(flattened)
+            if not finite_mask.any():
+                return np.nan
```
zechengz, line 61: I am not sure whether it's better to just raise error here? nan stats seem not useful?

zechengz (approved):

yiweny (commented):
~I think we can treat infinity as nans and drop the nans to calculate the column stats?~

yiweny (approved):

torch_frame/data/stats.py
weihua916: That's true. I think we can keep it for now, but I added a note to potentially error out in the future.

PR #197, opened 2023-11-09
Support saving and loading of `TensorFrame` with complex `stypes`

weihua916:
Made `MultiNestedTensor` into `dataclass` and wrote the serialization logic for saving and loading.
Supports loading and saving of `TensorFrame` with all complex stypes.

We can support `embedding` stype in the future soon.


torch_frame/utils/io.py
```diff
@@ -1,10 +1,48 @@
...
+from torch_frame.data import MultiNestedTensor, TensorFrame
 from torch_frame.data.stats import StatType
+from torch_frame.typing import TensorData
+
+
+def serialize_feat_dict(feat_dict: Dict[str, TensorData]) -> Dict[str, Any]:
+    feat_seriealized_dict = {}
+    for stype, feat in feat_dict.items():
+        if stype.use_multi_nested_tensor:
+            assert isinstance(feat, MultiNestedTensor)
```
zechengz, line 16: Maybe change to `_MultiTensor` as here it's also possible to be `MultiEmbeddingTensor`
```suggestion
            assert isinstance(feat, _MultiTensor)
```

torch_frame/utils/io.py
```diff
@@ -1,10 +1,48 @@
...
+            feat_seriealized = asdict(feat)
+        elif stype.use_dict_multi_nested_tensor:
+            feat_seriealized = {}
+            assert isinstance(feat, dict)
+            for name, f in feat.items():
+                assert isinstance(f, MultiNestedTensor)
```
zechengz, line 22: Same as above?

zechengz (approved):

torch_frame/utils/io.py
weihua916: For this one, we only have dict for multi-nested tensor, so we can keep it.

torch_frame/utils/io.py
weihua916: Goot point! Modified.

torch_frame/utils/io.py
```diff
@@ -1,10 +1,51 @@
...
+            assert isinstance(feat, Tensor)
+            feat_seriealized = feat
+        feat_seriealized_dict[stype] = feat_seriealized
+    return feat_seriealized_dict
+
+
+def deserialize_feat_dict(
+        feat_serialized_dict: Dict[str, Any]) -> Dict[str, TensorData]:
+    feat_dict = {}
+    for stype, feat_seriealized in feat_serialized_dict.items():
```
rusty1s, line 37: `serialized`

PR #198, opened 2023-11-10
Support `int` index in `MultiEmbeddingTensor`

akihironitta:
Part of #88.

```python
import torch
from torch_frame.data import MultiEmbeddingTensor
tensor_list = [
    torch.tensor([[0, 1, 2], [6, 7, 8]]),  # col1
    torch.tensor([[3, 4], [9, 10]]),       # col2
    torch.tensor([[5], [11]]),             # col3
]
met = MultiEmbeddingTensor.from_tensor_list(tensor_list)
print(met)
# MultiEmbeddingTensor(num_rows=2, num_cols=3, device='cpu')
print(met[0])
# MultiEmbeddingTensor(num_rows=1, num_cols=3, device='cpu')
print(met[0].values)
# tensor([[0, 1, 2, 3, 4, 5]])
print(met[0].offset)
# tensor([0, 3, 5, 6])
```

### DIFF
--- a/torch_frame/data/multi_embedding_tensor.py
+++ b/torch_frame/data/multi_embedding_tensor.py
@@ -206,32 +206,3 @@ class MultiEmbeddingTensor(_MultiTensor):
             # which is inconsistent with when dim=0
             offset = torch.LongTensor(offset_list)
             return MultiEmbeddingTensor(num_rows, num_cols, values, offset)
-
-    def _normalize_index(
-        self,
-        index: Union[int, Tensor],
-        dim: int,
-        is_slice_end: bool = False,
-    ) -> Union[int, Tensor]:
-        """Helper function to map negative indices to positive indices and
-        raise :obj:`IndexError` when necessary.
-
-        Args:
-            index: Union[int, Tensor]: Input :obj:`index` with potentially
-                negative elements.
-            is_slice_end (bool): Whether a given index (int) is slice or not.
-                If :obj:`True`, we have more lenient :obj:`IndexError`.
-                (default: :obj:`False`)
-        """
-        dim = self._normalize_dim(dim)
-        max_entries = self.num_rows if dim == 0 else self.num_cols
-        idx_name = "Row" if dim == 0 else "Col"
-        if isinstance(index, int):
-            if index < 0:
-                index = index + max_entries
-            if is_slice_end and index < 0 or index > max_entries:
-                raise IndexError(f"{idx_name} index out of bounds!")
-            elif (not is_slice_end) and (index < 0 or index >= max_entries):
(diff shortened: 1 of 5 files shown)


torch_frame/data/multi_embedding_tensor.py
```diff
@@ -196,3 +206,32 @@ def cat(
...
+        dim: int,
+        is_slice_end: bool = False,
+    ) -> Union[int, Tensor]:
+        """Helper function to map negative indices to positive indices and
+        raise :obj:`IndexError` when necessary.
+
+        Args:
+            index: Union[int, Tensor]: Input :obj:`index` with potentially
+                negative elements.
+            is_slice_end (bool): Whether a given index (int) is slice or not.
```
zechengz, line 222: Though not very sure, should it be `is slice end or not`?
```suggestion
            is_slice_end (bool): Whether a given index (int) is slice end or not.
```

zechengz (approved):

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -196,3 +206,32 @@ def cat(
             # which is inconsistent with when dim=0
             offset = torch.LongTensor(offset_list)
             return MultiEmbeddingTensor(num_rows, num_cols, values, offset)
+
+    def _normalize_index(
+        self,
+        index: Union[int, Tensor],
+        dim: int,
+        is_slice_end: bool = False,
```
weihua916, line 214: We already have `_normalize_index` in `MultiNestedTensor`. https://github.com/pyg-team/pytorch-frame/blob/master/torch_frame/data/multi_nested_tensor.py#L373-L409

What's the difference?

akihironitta: It's exactly the same. I'm moving it to the parent class

weihua916: Can we instead move `_normalize` to `_MultiTensor`, so that it can be used for both classes?

weihua916: Super, thanks.

torch_frame/data/multi_embedding_tensor.py
akihironitta: Done!

torch_frame/data/multi_embedding_tensor.py
akihironitta: Done! PTAL :)

torch_frame/data/multi_embedding_tensor.py
zechengz: Thanks!

weihua916 (approved):
Thanks!

PR #199, opened 2023-11-10
Support `list | torch.Tensor` index in `MultiEmbeddingTensor`

akihironitta:
Part of #88.

```python
import torch
from torch_frame.data import MultiEmbeddingTensor
tensor_list = [
    torch.tensor([[0, 1, 2], [6, 7, 8], [12, 13, 14]]),  # col1
    torch.tensor([[3, 4], [9, 10], [15, 16]]),           # col2
    torch.tensor([[5], [11], [17]]),                     # col3
]
met = MultiEmbeddingTensor.from_tensor_list(tensor_list)
print(met)
# MultiEmbeddingTensor(num_rows=3, num_cols=3, device='cpu')
print(met[[0, 2]])
# MultiEmbeddingTensor(num_rows=2, num_cols=3, device='cpu')
print(met[[0, 2]].values)
# tensor([[ 0,  1,  2,  3,  4,  5],
#         [12, 13, 14, 15, 16, 17]])

```

---

Follow-up tasks:
1. Add support for slicing in `MultiEmbeddingTensor`.
2. #194


torch_frame/data/multi_embedding_tensor.py
```diff
@@ -117,6 +94,70 @@ def from_tensor_list(
...
+        elif dim == 1:
+            return self._col_index_select(index)
+
+    def _row_index_select(self, index: Tensor) -> 'MultiEmbeddingTensor':
+        if index.numel() == 0:
+            return MultiEmbeddingTensor(
+                num_rows=0,
+                num_cols=self.num_cols,
+                values=torch.tensor([], device=self.device),
+                offset=torch.zeros(1, dtype=torch.long, device=self.device),
```
weihua916, line 152: the length of offset should be num_rows (0) * num_cols + 1 = 1.

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -38,30 +38,7 @@ class MultiEmbeddingTensor(_MultiTensor):
...
-            return MultiEmbeddingTensor(
-                num_rows=1,
-                num_cols=self.num_cols,
-                values=self.values[index].view(1, -1),
-                offset=self.offset,
-            )
-
-        # TODO(akihironitta): Support more index types
-        raise NotImplementedError
+        assert len(self.offset) == min(1, self.num_rows) * self.num_cols + 1
```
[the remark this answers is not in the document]
akihironitta: Ah yes, mine is wrong. Is your suggestion still correct though? Looking at what we do in `MultiNestedTensor`:
https://github.com/pyg-team/pytorch-frame/blob/e8a11d43746d6b69b97e1d3646451d8a88a9ab88/torch_frame/data/multi_nested_tensor.py#L287-L293
I thought an empty `MultiEmbeddingTensor`'s `offset` should be `torch.tensor([0])` and this assertion should be:

```python
assert len(self.offset) == min(1, max(1, self.num_rows)) * self.num_cols + 1
```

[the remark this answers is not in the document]
akihironitta: maybe this is easier to read:
```python
if self.num_rows == 0:
    assert len(self.offset) == 1
else:
    assert len(self.offset) == self.num_cols + 1
```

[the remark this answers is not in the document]
weihua916: I think mine should be correct. 0 * num_cols + 1 = 0, right?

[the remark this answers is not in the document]
weihua916: We have num_rows * num_cols entries, and the number of elements in offset should be the number of entries + 1.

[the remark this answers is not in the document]
weihua916: Actually, I think the offset should be `assert len(self.offset) == self.num_cols + 1`
Even when the number of rows is 0, I think we can still store the column offset to prepare for concatenation.

torch_frame/data/multi_embedding_tensor.py
weihua916: Nvm, I was wrong :( Since we are implementing `MultiEmbeddingTensor` (not nested tensor), we only need to store `num_cols + 1` offset.

torch_frame/data/multi_embedding_tensor.py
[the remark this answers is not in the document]
akihironitta: > 0 * num_cols + 1 = 0, right?

Yes, but it'll be wrong when `num_rows>1` because, in `MultiEmbeddingTensor`, the size of `offset` stays the same at `num_cols+1` when `num_rows` gets bigger, no? I might be missing something :)

torch_frame/data/multi_embedding_tensor.py
weihua916: I don't think we need special handling of `index.numel() == 0` case here.

akihironitta: Done! Please have another look

test/data/test_multi_embedding_tensor.py
```diff
@@ -117,13 +117,46 @@ def test_index():
...
+        met_indexed = met[torch.tensor(index, dtype=torch.long)]
+        assert isinstance(met_indexed, MultiEmbeddingTensor)
+        assert met_indexed.shape[0] == len(index)
+        assert met_indexed.shape[1] == num_cols
+        for i, idx in enumerate(index):
+            for j in range(num_cols):
+                assert torch.allclose(
+                    tensor_list[j][idx],
+                    met_indexed[i, j],
+                )
```
weihua916, line 152: Nit: can we merge these tests? A lot of code repetition.

weihua916:
```python
for index_type in ['list', 'tensor']:
    if index_type == 'tensor':
         index = torch.tensor(index, dtype=torch.long)
    met_indexed = met[index]
    ....
```

weihua916 (approved):
LGTM, final minor comment

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -117,6 +94,64 @@ def from_tensor_list(
...
+    def __getitem__(
+        self,
+        index: Any,
+    ) -> Union["MultiEmbeddingTensor", Tensor]:
+        if isinstance(index, tuple) and len(index) == 2 and isinstance(
+                index[0], int) and isinstance(index[1], int):
+            i = self._normalize_index(index[0], dim=0)
+            j = self._normalize_index(index[1], dim=1)
+            return self.values[i, self.offset[j]:self.offset[j + 1]]
+        if isinstance(index, int):
```
zechengz, line 106: nit: change if to elif?

zechengz (approved):

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -117,6 +94,64 @@ def from_tensor_list(
         offset = torch.LongTensor(offset_list)
         return cls(num_rows, num_cols, values, offset)
 
+    def __getitem__(
+        self,
+        index: Any,
+    ) -> Union["MultiEmbeddingTensor", Tensor]:
+        if isinstance(index, tuple) and len(index) == 2 and isinstance(
```
yiweny, line 101: If the tuple is `(1, '4')` then this function will raise a `NotImplementedError`, right?

yiweny: Can you change it to raise a different error if the format of the Tuple is wrong? Thanks!

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -117,6 +94,64 @@ def from_tensor_list(
...
+            return self.values[i, self.offset[j]:self.offset[j + 1]]
+        if isinstance(index, int):
+            index = self._normalize_index(index, dim=0)
+            return MultiEmbeddingTensor(
+                num_rows=1,
+                num_cols=self.num_cols,
+                values=self.values[index].view(1, -1),
+                offset=self.offset,
+            )
+        if isinstance(index, Tensor) and index.ndim == 1:
```
yiweny, line 114: I think we need to check the device here and change the device of the index if it's different from `MultiEmbeddingTensor`. Do you mind also adding gpu testing? Thanks

yiweny (commented):
Appreciate the work~ Left a few comments.

torch_frame/data/multi_embedding_tensor.py
weihua916: I think it is fine for now. The feature is still being developed, and when we get feature-complete, let's raise an informative error, like xxx indexing is not supported.

torch_frame/data/multi_embedding_tensor.py
weihua916: Just to be clear, by "GPU testing", we do not mean adding a new GPU unit test. Just running an adhoc test on your local GPU machine is enough for now.

torch_frame/data/multi_embedding_tensor.py
weihua916: Excellent

torch_frame/data/multi_embedding_tensor.py
akihironitta: To make the behaviour consistent with PyTorch which accepts both cpu/cuda tensors as index, I'll just keep this as is, but I'll enable gpu testing for this test case.

PR #200, opened 2023-11-10
Update dev dependency

akihironitta:
I've just realised that we were installing https://pypi.org/project/torch-frame/ as one of our dependencies.

Please check and uninstall it from your dev environment:
```bash
pip uninstall torch-frame
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,7 +40,7 @@ test=[
     "pytest-cov",
 ]
 dev=[
-    "torch_frame[test]",
+    "pytorch_frame[test]",
     "pre-commit",
 ]
 full=[


pyproject.toml
```diff
@@ -40,7 +40,7 @@ test=[
     "pytest-cov",
 ]
 dev=[
-    "torch_frame[test]",
+    "pytorch_frame[test]",
```
akihironitta, line 43: Just for the note,

```
[tool.flit.module]
name="torch_frame"
```

is fine as we're supposed to fill it with import name but not with package name on PyPI.

akihironitta: https://flit.pypa.io/en/stable/pyproject_toml.html#module-section

weihua916 (approved):

PR #201, opened 2023-11-10
add voyageai example

yiweny:


examples/llm_embedding.py
```diff
@@ -81,15 +85,38 @@ def __call__(self, sentences: List[str]) -> Tensor:
...
+
+    def __init__(self, model: str = 'voyage-01'):
+        # Please run `pip install voyageai` to install the package
+        self.model = model
+
+    def __call__(self, sentences: List[str]) -> Tensor:
+        import voyageai
+        voyageai.api_key = api_key
+        from voyageai import get_embeddings
+        items = get_embeddings(sentences, model=self.model)
```
weihua916, line 100: Just curious, what's the type of `items`? Can we have the same type annotation for the other API embeddings as well?

yiweny: `List[List[Float]]`

weihua916 (approved):
Thanks! How does it perform?

yiweny:
> Thanks! How does it perform?

still running experiments.

Issue #154, opened 2023-10-29
Add overwrite logic for dataset materialization


yiweny:
Currently, `dataset.materialize()` supports `path` argument where it loads a cached object. 
However, if any column changes, for the dataset, or if we are using a different transformer model, the default behavior is to read from the cached object rather than rematerialize. 

I think it's necessary to add logic to overwrite the cached object if the dataset changes. Currently, there's not an option to do so--you have to manually delete the cached object.

weihua916:
If we do not specify the `path`, we will automatically over-write the cached object.

zechengz:
@weihua916 Right, but it will not replace the old cache. User need to manually remove old cache.

weihua916:
Manually removing old cache seems fine to me.

yiweny:
@weihua916 One thing I found a bit cumbersome is that when I changed the text embedder, I need to delete the old cache. It's less intuitive to users that the embedding takes place at materialization stage. And if the size of the embedders are the same, then the users would easily thought that they don't need to re-materialize the data and nothing will pop a warning.

weihua916:
I think one way to resolve this issue is to make [this line](https://github.com/pyg-team/pytorch-frame/blob/master/examples/llm_embedding.py#L101) to be dependent on `args.service`.

```python
dataset.materialize(path=osp.join(path, f'{args.service}_data.pt'))
```

How about we add this to our example?

PR #204, opened 2023-11-10
Config specific `path` during dataset materialization

weihua916:
Fixes https://github.com/pyg-team/pytorch-frame/issues/154 by using different paths for caching different tensor frame objects.


zechengz (approved):
Thanks!

PR #208, opened 2023-11-13
Add `multi-categorical` to `MultimodalTextBenchmark`

zechengz:
- Add `multi-categorical` to `MultimodalTextBenchmark`
- Add `pyarrow` to the dependencies as need it to read parquet for `MultimodalTextBenchmark`

### DIFF
--- a/torch_frame/datasets/multimodal_text_benchmark.py
+++ b/torch_frame/datasets/multimodal_text_benchmark.py
@@ -215,6 +215,16 @@ class MultimodalTextBenchmark(torch_frame.data.Dataset):
         'mercari_price_suggestion100K': ['train', 'test'],
     }
 
+    _dataset_col_to_sep = {
+        'data_scientist_salary': ', ',
+        'imdb_genre_prediction': ', ',
+        'california_house_price': ', ',
+        'melbourne_airbnb': ', ',
+        'ae_price_prediction': ', ',
+        'kick_starter_funding': ', ',
+        'mercari_price_suggestion100K': '/',
+    }
+
     _dataset_stype_to_col = {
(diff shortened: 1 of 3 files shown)


torch_frame/datasets/multimodal_text_benchmark.py
```diff
@@ -482,7 +498,11 @@ def __init__(self, root: str, name: str,
                     col_to_stype[col] = self.text_stype
                 else:
                     col_to_stype[col] = stype
+
+        df = self._pre_transform(df=df, target_col=target_col)
+        col_to_sep = self._dataset_col_to_sep.get(name, '')
```
weihua916, line 503: nit: why do we return the empty string by default? When does that happen?

weihua916 (approved):
Clean, thanks. Shall we redo the experiments (and update the results in `examples/`) based on the update?

yiweny (approved):

torch_frame/datasets/multimodal_text_benchmark.py
zechengz: When make `col_to_sep`  as none it will raise an error. `col_to_sep` now has some default value `", "` in the dataset.py

weihua916: Why don't you raise an error if `name` is not in `self._dataset_col_to_sep`?

zechengz: This means that there is no multicategorical for that dataset `name`

weihua916: Got it. LGTM as it is now. Thanks for clarifying.

PR #211, opened 2023-11-13
Update the transformers text example

zechengz:
Update the transformers text example with
- Fix the data save error when model name contains `/`
- Add mixed precision training, which saves memory and makes the training faster. But unfortunately for some datasets, marked as `N/A` in the comments, may need to tune or adjust the `scaler` scale or clip the gradient norm (https://pytorch.org/docs/stable/generated/torch.nn.utils.clip_grad_norm_.html) to get reasonable results
- Count and print number of finetune parameters for the text encoder.

### DIFF
--- a/examples/transformers_text.py
+++ b/examples/transformers_text.py
@@ -25,6 +25,8 @@ from torch_frame.nn import (
     LinearModelEncoder,
 )
 from torch_frame.typing import TensorData
+from accelerate import init_empty_weights
+from accelerate.utils import BnbQuantizationConfig, load_and_quantize_model
 
 # Text Embedded
 # all-distilroberta-v1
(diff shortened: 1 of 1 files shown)


examples/transformers_text.py
```diff
@@ -54,6 +72,7 @@
 parser.add_argument('--seed', type=int, default=0)
 parser.add_argument('--finetune', action='store_true')
 parser.add_argument('--lora', action='store_true')
+parser.add_argument('--mixed', action='store_true')
```
weihua916, line 75: rename to `mixed_precision_training`

weihua916 (approved):

weihua916:
Personally, I'd prefer mixed precision not to be merged into `master` to avoid clutter on our example script. I am more comfortable with merging the first and third points in this PR, and keep experimenting with mixed precision training until we are confident that it works.

zechengz:
@weihua916 Sure!

akihironitta:
@zechengz In addition to `bf16`, you could also try disabling casting in particular operations that lead to the instability issue.
https://discuss.pytorch.org/t/disabling-mixed-precsion-in-my-own-layers/97190

zechengz:
@akihironitta Thanks, I will have a try later.

Issue #206, opened 2023-11-10
Allow different `text_embedder_cfg` for each column


toenshoff:
Currently, the `Dataset` class expects a single `text_embedder_cfg` object that is used globally to embed all text embedding columns.

This may not be desirable in cases where different pre-trained models should be applied to different columns.
An important example would be multi-lingual data where different text columns have different languages, say English and Chinese.

I think it would be great if one can also optionally pass a dictionary with columns-wise configs, similar to the `col_to_sep` argument.
I would gladly add this feature if it is considered helpful.

Thanks for this project by the way, it will be very useful for many datasets.

toenshoff:
I see, that makes sense. 

Would it be reasonable to add the embedding dimension as an attribute to `StatType`? 
It may be cumbersome to manage this information separately from other column statistics,  especially when there are multiple embedded columns with varying dimensions.

weihua916:
Yes, that makes sense. We were working towards that direction, but the challenge was that column stats are computed purely based on the pandas columns (`ser` [here](https://github.com/pyg-team/pytorch-frame/blob/master/torch_frame/data/dataset.py#L430-L431)). We cannot obtain the embedding dimensionality information until we actually apply the embedder, which is why we let our users manually specify the dimensionality.

Do you have any other ideas?

toenshoff:
I agree that this is a little tricky. 

The simplest solution may be to add a `dim` attribute to `TextEmbedderConfig` which enables the user to specify the embedding dimension in the `text_embedder_cfg`  objects that are passed. This could then be passed on as a new argument to `compute_col_stats` [here](https://github.com/pyg-team/pytorch-frame/blob/e8a11d43746d6b69b97e1d3646451d8a88a9ab88/torch_frame/data/dataset.py#L431C1-L431C64) just like the optional separators already are. As a downside, this would clutter the signature of `compute_col_stats` and a user may provide the wrong embedding dimension.

A second solution would be to add a new method that updates existing `col_stats` given a `TensorFrame` object. So something like:
```python
update_col_stats(stats: Dict[StatType, Any], tensor_frame: TensorFrame) -> Dict[StatType, Any]
```
This could be called for the stats of each column after materializing the frame to add further information such as embedding dimensions.

Let me know if these suggestions make any sense.

weihua916:
[...]

PR #217, opened 2023-11-15
Support slicing in `MultiEmbeddingTensor`

akihironitta:
Part of #88. Unblocks #194 and #206.

### DIFF
--- a/test/data/test_multi_embedding_tensor.py
+++ b/test/data/test_multi_embedding_tensor.py
@@ -21,12 +21,11 @@ def assert_equal(
             assert torch.allclose(tensor_list[j][i], met[i, j])
 
 
-# FIXME: Check again when using this to test something
 def row_select(
     tensor_list: List[torch.Tensor],
     index: Union[List[int], slice],
 ) -> List[torch.Tensor]:
-    """Selects rows from a list of tensors.
+    """Selects rows from a list of column tensors.
 
     Args:
         tensor_list (list[torch.Tensor]): a list of tensors of size
@@ -294,6 +293,12 @@ def test_narrow(device):
     )
     assert_equal(tensor_list[2:2 + 4], met.narrow(1, 2, 4))
     assert_equal(row_select(tensor_list, slice(2, 2 + 4)), met.narrow(0, 2, 4))
+    empty_met = met.narrow(0, 2, 0)
+    assert empty_met.shape[0] == 0
+    assert empty_met.shape[1] == num_cols
+    empty_met = met.narrow(1, 2, 0)
+    assert empty_met.shape[0] == num_rows
+    assert empty_met.shape[1] == 0
 
 
 def test_clone():
--- a/test/data/test_multi_embedding_tensor.py
+++ b/test/data/test_multi_embedding_tensor.py
@@ -30,6 +30,8 @@ def row_select(
     Args:
         tensor_list (list[torch.Tensor]): a list of tensors of size
             [num_rows, dim_emb_j].
+        index (Union[List[int], slice]): a list of row indices or a slice to
+            apply to each tensor in tensor_list.
 
     Returns:
         List[torch.Tensor]: new_tensor_list is a list of tensors of size
--- a/test/data/test_multi_embedding_tensor.py
+++ b/test/data/test_multi_embedding_tensor.py
@@ -169,7 +169,6 @@ def test_index(device):
                     )
 
 
-# FIXME: Merge this test case with test_index
 @withCUDA
 def test_index_range(device):
     num_rows = 8
@@ -190,7 +189,6 @@ def test_index_range(device):
                 assert torch.allclose(tensor_list[j][i], met_indexed[idx, j])
 
 
-# FIXME: Merge this test case with test_index
 @withCUDA
 def test_index_slice(device):
     num_rows = 8
@@ -213,7 +211,6 @@ def test_index_slice(device):
     assert empty_met.shape[1] == num_cols
 
 
-# FIXME: Merge this test case with test_index
 @withCUDA
 def test_index_slice_slice(device):
     num_rows = 8
@@ -236,7 +233,6 @@ def test_index_slice_slice(device):
     assert empty_met.shape[1] == 0
 
 
-# FIXME: Merge this test case with test_index
 @withCUDA
 def test_index_slice_list(device):
     num_rows = 8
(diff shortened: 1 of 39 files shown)


torch_frame/data/multi_embedding_tensor.py
```diff
@@ -98,29 +101,159 @@ def __getitem__(
...
-            index = self._normalize_index(index, dim=0)
-            return MultiEmbeddingTensor(
-                num_rows=1,
-                num_cols=self.num_cols,
-                values=self.values[index].view(1, -1),
-                offset=self.offset,
-            )
+        if isinstance(index, tuple):
+            assert len(index) == 2
+            if isinstance(index[0], int) and isinstance(index[1], int):
```
zechengz, line 106: nit
```suggestion
            if all(isinstance(x, int) for x in index):
```

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -98,29 +101,159 @@ def __getitem__(
...
+                values=torch.tensor([], device=self.device),
+                offset=torch.tensor([0], device=self.device),
+            )
+
+        elif dim == 0:
+            return MultiEmbeddingTensor(
+                num_rows=length,
+                num_cols=self.num_cols,
+                values=self.values[start:start + length],
+                offset=self.offset,
```
zechengz, line 218: Should we also change the offset here?

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -98,29 +101,159 @@ def __getitem__(
...
+                out = self
+                for dim, idx in enumerate(index):
+                    out = out.select(idx, dim=dim)
+                return out
+        else:
+            return self.select(index, dim=0)
+
+    def select(
+        self,
+        index: Union[int, Tensor, List[int], slice, range],
```
yiweny, line 120: Should `Tuples` be supported?

test/data/test_multi_embedding_tensor.py
```diff
@@ -21,6 +21,29 @@ def assert_equal(
...
+    tensor_list: List[torch.Tensor],
+    index: Union[List[int], slice],
+) -> List[torch.Tensor]:
+    """Selects rows from a list of column tensors.
+
+    Args:
+        tensor_list (list[torch.Tensor]): a list of tensors of size
+            [num_rows, dim_emb_j].
+        index (Union[List[int], slice]): a list of row indices or a slice to
+            apply to each tensor in tensor_list.
```
weihua916, line 34: A list of...

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -38,7 +38,10 @@ class MultiEmbeddingTensor(_MultiTensor):
     """
     def validate(self):
         assert self.offset[0] == 0
-        assert len(self.offset) == self.num_cols + 1
+        assert len(
+            self.offset) == self.num_cols + 1 or self.values.numel() == 0
```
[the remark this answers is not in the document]
weihua916: Why can't we let `len(self.offset) == self.num_cols + 1` for regardless of whether the tensor is empty?

torch_frame/data/multi_embedding_tensor.py
weihua916: I don't think row indexing (`dim=0`) operation should change the offset, since `offset` here purely specifies the column slicing information.

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -98,29 +101,159 @@ def __getitem__(
...
+            end_idx: int = self._normalize_index(
+                index.stop or num_data,
+                dim=dim,
+                is_slice_end=True,
+            )
+            return self.narrow(
+                dim=dim,
+                start=start_idx,
+                length=end_idx - start_idx,
+            )
```
weihua916, line 255: Do you think `_slice`, `narrow`, `select`, `index_select` can be shared with `MultiNestedTensor` (in other words, we put them in the parent `_MultiTensor` class)? Here we only need to implement `_row_index_select`, `_col_index_select` and so on.

weihua916 (commented):
LGTM in general, thanks!

i think we can put some general methods to parent class `_MultiTensor` to allow more code sharing with `MultiNestedTensor`.

torch_frame/data/multi_embedding_tensor.py
akihironitta: Sounds good. I'll update the PR accordingly

torch_frame/data/multi_embedding_tensor.py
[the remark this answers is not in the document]
akihironitta: Because we still keep the size (incl num_cols) for an empty instance. What would you suggest in this case?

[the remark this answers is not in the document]
weihua916: For empty instance, we can still have positive num_cols and associated offset?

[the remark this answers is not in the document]
akihironitta: Ah I might be misunderstanding something. I'll double check this once I'm back at computer. Thank you for your thorough review as always!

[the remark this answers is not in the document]
akihironitta: Fixed!

torch_frame/data/multi_embedding_tensor.py
akihironitta: Actually, could I follow up on the refactor in a separate PR to first unblock #194 and #206? I'd like to keep this PR easy to review and not touch `MultiNestedTensor` if you're fine.

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -152,6 +285,60 @@ def _row_index_select(self, index: Tensor) -> 'MultiEmbeddingTensor':
...
+        torch.cumsum(new_col_dims, dim=0, out=offset[1:])
+        value_index = [
+            torch.arange(
+                self.offset[i],
+                self.offset[i + 1],
+                dtype=torch.long,
+                device=self.device,
+            ) for i in index
+        ]
+        value_index = torch.cat(value_index, dim=0)
```
weihua916, line 312: Can you try using `batched_arange` to avoid for-loop? See my usage in `MultiNestedTensor`.

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -152,6 +285,60 @@ def _row_index_select(self, index: Tensor) -> 'MultiEmbeddingTensor':
...
+                num_cols=self.num_cols,
+                values=self.values[index].view(1, -1),
+                offset=self.offset,
+            )
+        elif dim == 1:
+            return MultiEmbeddingTensor(
+                num_rows=self.num_rows,
+                num_cols=1,
+                values=self.values[:,
+                                   self.offset[index]:self.offset[index + 1]],
```
weihua916, line 338: super nit: a bit more beautiful indent?

torch_frame/data/multi_embedding_tensor.py
weihua916: Sounds good. Thanks.

weihua916 (approved):
LGTM after avoiding for-loop via `batched_arange` (see [here](https://github.com/pyg-team/pytorch-frame/blob/master/torch_frame/data/multi_nested_tensor.py#L492-L518)). Thank you.

PR #218, opened 2023-11-15
Make `Trompt` output 2d embeddings in `forward`.

weihua916:
so that the behavior of all tensor frame models is aligned.
Instead we add `forward_stacked` that outputs stacked embeddings for layer-wise loss computation.


yiweny (approved):
LGTM

Issue #216, opened 2023-11-15
Clean up `TextTokenizer` logic


weihua916:
Currently, we are doing hacking around [here](https://github.com/pyg-team/pytorch-frame/blob/master/torch_frame/data/mapper.py#L296-L305). We should remove this hack.

zechengz:
Actually I take a second look, as shown in the text example https://github.com/pyg-team/pytorch-frame/blob/b8d8d56a1c240e39e0ae47be06c0306776fb9c07/examples/transformers_text.py#L102-L104
where each tensor (values of `inputs`) is a 2 dimensional tensor if tokenizer receives a list of sentences but it will return a 1 dimensional tensor if tokenizer receives just one sentence. So, I think we should support both `dim=1` and `dim=2` tensors. WDYT?

weihua916:
I think we can decompose this 2d tensor into 1d tensor using for-loop. We do for-loop anyway. I think it's better to be transparent and do for-loop here rather than inside our package. What do you think?

zechengz:
Just to double check you mean we should do for loop in the text tokenizer such as here https://github.com/pyg-team/pytorch-frame/blob/b8d8d56a1c240e39e0ae47be06c0306776fb9c07/examples/transformers_text.py#L102-L104 instead of the mapper? IMO it's natural to just return tokens / masks in 2D tensor and we can help to handle in the mapper.

weihua916:
Yes, for-loop here instead of inside `TensorMapper`.

weihua916:
We aligned that we will define
```python
TextTokenizationOutputs = Union[List[Dict[str, Tensor]], Dict[str, Tensor]]
```
and implement a logic for `Dict[str, Tensor]` accordingly.

PR #219, opened 2023-11-16
Clean up TextTokenizer

zechengz:
Fix https://github.com/pyg-team/pytorch-frame/issues/216

### DIFF
--- a/test/datasets/test_fake_dataset.py
+++ b/test/datasets/test_fake_dataset.py
@@ -11,7 +11,8 @@ from torch_frame.testing.text_tokenizer import WhiteSpaceHashTokenizer
 
 
 @pytest.mark.parametrize('with_nan', [True, False])
-def test_fake_dataset(with_nan):
+@pytest.mark.parametrize('tokenize_with_batched', [True, False])
+def test_fake_dataset(with_nan, tokenize_with_batched):
     num_rows = 20
     out_channels = 10
     dataset = FakeDataset(
@@ -29,7 +30,8 @@ def test_fake_dataset(with_nan):
         text_embedder_cfg=TextEmbedderConfig(
             text_embedder=HashTextEmbedder(out_channels), batch_size=None),
         text_tokenizer_cfg=TextTokenizerConfig(
-            text_tokenizer=WhiteSpaceHashTokenizer(num_hash_bins=12),
+            text_tokenizer=WhiteSpaceHashTokenizer(
+                num_hash_bins=12, batched=tokenize_with_batched),
             batch_size=5),
     )
     assert str(dataset) == 'FakeDataset()'
--- a/test/datasets/test_fake_dataset.py
+++ b/test/datasets/test_fake_dataset.py
@@ -11,8 +11,9 @@ from torch_frame.testing.text_tokenizer import WhiteSpaceHashTokenizer
 
 
 @pytest.mark.parametrize('with_nan', [True, False])
-@pytest.mark.parametrize('tokenize_with_batched', [True, False])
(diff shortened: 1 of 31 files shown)


test/datasets/test_fake_dataset.py
```diff
@@ -11,7 +11,8 @@
 
 
 @pytest.mark.parametrize('with_nan', [True, False])
-def test_fake_dataset(with_nan):
+@pytest.mark.parametrize('tokenize_with_batched', [True, False])
```
weihua916, line 14: tokenize_with_batch

torch_frame/typing.py
```diff
@@ -63,5 +63,6 @@ def is_numerical_strategy(self):
 
 IndexSelectType = Union[int, List[int], range, slice, Tensor]
 ColumnSelectType = Union[str, List[str]]
-TextTokenizationOutputs = List[Dict[str, Tensor]]
+TextTokenizationOutputs = Union[List[Mapping[str, Tensor]], Mapping[str,
+                                                                    Tensor]]
```
weihua916, line 67: super nit: awkward line breaking. Maybe add parenthesis.

torch_frame/data/mapper.py
```diff
@@ -341,15 +343,19 @@ def forward(
...
-        for key in tokenized_list[0]:
+            if isinstance(tokenized_batch, Mapping):
+                tokenized_outputs.append(tokenized_batch)
+            else:
+                tokenized_outputs.extend(tokenized_batch)
+
+        for key in tokenized_outputs[0]:
             xs = []
-            for item in tokenized_list:
+            for item in tokenized_outputs:
```
weihua916, line 358: Can we branch 
```python
                if item[key].ndim == 1:
                    xs.append([item[key]])
                elif item[key].ndim == 2:
                    for row in item[key]:
                        xs.append([row])
```
based on 
```python
if isinstance(tokenized_batch, Mapping):
```
and then assert on ndim?

weihua916, line 358: `for tokenized_batch in tokenized_outputs`
to align with your previous for-loop

zechengz: Do you mean if the `tokenized_batch` is `Mapping`, tensors should be dim=2?

torch_frame/data/mapper.py
```diff
@@ -328,43 +328,38 @@ def forward(
...
+                for tokenized_batch in tokenized_outputs:
+                    xs.extend([row] for row in tokenized_batch[key])
+                feat_dict[key] = MultiNestedTensor.from_tensor_mat(xs)
+        else:
+            keys = tokenized_outputs[0][0].keys()
+            for key in keys:
+                xs = []
+                for tokenized_batch in tokenized_outputs:
+                    xs.extend([item[key]] for item in tokenized_batch)
+                feat_dict[key] = MultiNestedTensor.from_tensor_mat(xs)
```
weihua916, line 362: nit: the following can be put outside of if-else statement.
```python
feat_dict[key] = MultiNestedTensor.from_tensor_mat(xs)
```

weihua916, line 362: Can we raise error on `ndim` of `tokenized_batch[key]`?

zechengz: I think for each key we need to call `from_tensor_mat` and each type of `tokenized_outputs` (mapping or list) need to get `keys` in a different way. So I think it still need to be put inside the if else.

weihua916: Oh I see. Thanks for clarifying.

weihua916 (approved):
Very clean, thanks!

PR #225, opened 2023-11-17
add TimestampEncoder

yiweny:


torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -15,6 +22,14 @@
 
 from ..utils.init import attenuated_kaiming_uniform_
 
+NUM_DAYS_PER_MONTH = 31
+NUM_MONTHS_PER_YEAR = 12
+NUM_DAYS_PER_MONTH = 31
```
zechengz, line 27: Duplicated? Also some months don't have 30 days right?

yiweny: ohh good catch. Yeah but I want to keep the scale between 0 and 1, so I divide by the largest possible number of days

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -685,3 +699,67 @@ def encode_forward(self, feat: TensorData) -> Tensor:
         # -> [batch_size, num_cols, out_channels]
         x = x_lin + self.bias
         return x
+
+
+class Time2VecEncoder(StypeEncoder):
```
weihua916, line 704: TimestampEncoder

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -685,3 +699,67 @@ def encode_forward(self, feat: TensorData) -> Tensor:
...
+        feat[:, :, 2] /= NUM_DAYS_PER_MONTH
+        feat[:, :, 3] /= NUM_DAYS_PER_WEEK
+        feat[:, :, 4] /= NUM_HOURS_PER_DAY
+        feat[:, :, 5] /= NUM_MINUTES_PER_HOUR
+        feat[:, :, 6] /= NUM_SECONDS_PER_MINUTE
+        linear_output = self.linear_1(feat[:, :, 0].unsqueeze(-1))
+        periodic_output = self.periodic(feat[:, :, 1:])
+        periodic_output = torch.sin(periodic_output)
+        output = torch.cat((linear_output, periodic_output), dim=-1)
+        return self.linear_2(output)
```
weihua916, line 765: apply sin cos with different frequencies first, then linear. Do you mind I write this part later tomorrow?

test/nn/encoder/test_stype_encoder.py
```diff
@@ -148,6 +149,35 @@ def test_multicategorical_feature_encoder(encoder_cls_kwargs):
...
+
+    encoder = encoder_cls_kwargs[0](
+        8,
+        stats_list=stats_list,
+        stype=stype.timestamp,
+        **encoder_cls_kwargs[1],
+    )
+    feat_timestamp = tensor_frame.feat_dict[stype.timestamp]
+    x = encoder(feat_timestamp)
+    assert x.shape == (feat_timestamp.size(0), feat_timestamp.size(1), 8)
```
weihua916, line 178: Can we add a final integration test on `test_stypewise_encoder.py`?

weihua916 (commented):
Thanks! Left some comments!

torch_frame/nn/encoder/stype_encoder.py
yiweny: Let’s discuss tomorrow offline!

test/nn/encoder/test_stype_encoder.py
yiweny: I need to figure out a good na strategy for timestamps first. Let’s discuss offline tomorrow!

torch_frame/nn/encoder/stype_encoder.py
zechengz: I think we need to have something like `torch.sin(feat * 2 * torch.pi)` before feeding it to the `periodic` layer

torch_frame/data/mapper.py
```diff
@@ -259,7 +259,8 @@ def forward(
             torch.from_numpy(
                 ser.dt.second.values).to(device=device).unsqueeze(1)
         ]
-        return torch.stack(tensors).permute(1, 2, 0).to(torch.float32)
+        return torch.stack(tensors).permute(1, 2,
+                                            0).squeeze(1).to(torch.float32)
```
rusty1s, line 263: Why convert to float?

torch_frame/nn/encoder/stype_encoder.py
rusty1s: Might be good to add this as a comment, potentially with a TODO to revisit this.

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -685,3 +699,67 @@ def encode_forward(self, feat: TensorData) -> Tensor:
...
+
+class Time2VecEncoder(StypeEncoder):
+    r"""Time2VecEncoder for encoding timestamp stype. The input tensor is
+    expected to have 7 columns representing year, month, day, day of week,
+    hour, minute, and second. The encoding includes both a linear component
+    (for the year) and periodic components (for month, day, day of week,
+    hour, minute, second). The linear component applies
+    :class:`torch.nn.Linear` on the year and the periodic components apply
+    sin function on the rest of the columns. And then a
+    :class:`torch.nn.Linear` is applied to the concatenated components.
```
rusty1s, line 712: Don't get that. Why not apply periodic components on subset, merge with year, and then apply a single `Linear` layer?

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -685,3 +699,67 @@ def encode_forward(self, feat: TensorData) -> Tensor:
...
+        self.periodic = Linear(stype.timestamp.timestamp_num_cols - 1,
+                               stype.timestamp.timestamp_num_cols - 1)
+        self.min_year = torch.tensor([
+            self.stats_list[i][StatType.YEAR_RANGE][0]
+            for i in range(len(self.stats_list))
+        ])
+        self.max_year = torch.tensor([
+            self.stats_list[i][StatType.YEAR_RANGE][1]
+            for i in range(len(self.stats_list))
+        ])
```
rusty1s, line 739: should be buffers.

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -685,3 +699,67 @@ def encode_forward(self, feat: TensorData) -> Tensor:
...
+    def reset_parameters(self):
+        super().reset_parameters()
+        self.linear_1.reset_parameters()
+        self.linear_2.reset_parameters()
+
+    def encode_forward(self, feat: Tensor) -> Tensor:
+        device = feat.device
+        feat[:, :, 0] = (feat[:, :, 0] - self.min_year.to(device)) / (
+            self.max_year.to(device) - self.min_year.to(device) + 1)
+        feat[:, :, 1] -= 1
```
rusty1s, line 753: Should be done in materialization.

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -685,3 +699,67 @@ def encode_forward(self, feat: TensorData) -> Tensor:
...
+        feat[:, :, 1] /= NUM_MONTHS_PER_YEAR
+        feat[:, :, 2] -= 1
```
rusty1s, line 755: Here as well.

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -685,3 +699,67 @@ def encode_forward(self, feat: TensorData) -> Tensor:
...
+        feat[:, :, 2] /= NUM_DAYS_PER_MONTH
+        feat[:, :, 3] /= NUM_DAYS_PER_WEEK
+        feat[:, :, 4] /= NUM_HOURS_PER_DAY
+        feat[:, :, 5] /= NUM_MINUTES_PER_HOUR
+        feat[:, :, 6] /= NUM_SECONDS_PER_MINUTE
```
rusty1s, line 760: Would be better to somewhere define which index refers to which date information, e.g.:
```python
TIME_TO_INDEX + {
    'YEAR': 0,
    'WEEK': 1,
}
```

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -685,3 +699,67 @@ def encode_forward(self, feat: TensorData) -> Tensor:
...
+            self.stats_list[i][StatType.YEAR_RANGE][1]
+            for i in range(len(self.stats_list))
+        ])
+        self.linear_2 = Linear(stype.timestamp.timestamp_num_cols,
+                               self.out_channels)
+        self.reset_parameters()
+
+    def reset_parameters(self):
+        super().reset_parameters()
+        self.linear_1.reset_parameters()
```
rusty1s, line 746: Reset `periodic`.

torch_frame/data/mapper.py
yiweny: I think the values by default is `float64`

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -685,3 +724,73 @@ def encode_forward(self, feat: TensorData) -> Tensor:
...
+            self.stats_list[i][StatType.YEAR_RANGE][0]
+            for i in range(len(self.stats_list))
+        ])
+        self.register_buffer("min_year", min_year)
+        max_year = torch.tensor([
+            self.stats_list[i][StatType.YEAR_RANGE][1]
+            for i in range(len(self.stats_list))
+        ])
+        self.register_buffer("max_year", max_year)
+        self.mult_constants = [torch.pi, torch.pi * 2, torch.pi * 3]
```
zechengz, line 765: Just curious why we also multiply with `torch.pi` and `torch.pi * 3`

torch_frame/typing.py
```diff
@@ -44,6 +44,9 @@ class NAStrategy(Enum):
     MEAN = 'mean'
     MOST_FREQUENT = 'most_frequent'
     ZEROS = 'zeros'
+    MOST_RECENT_TIMESTAMP = 'most_recent_timestamp'
```
zechengz, line 47: LATEST_TIMESTAMP?

torch_frame/typing.py
```diff
@@ -44,6 +44,9 @@ class NAStrategy(Enum):
...
+    LEAST_RECENT_TIMESTAMP = 'least_recent_timestamp'
```
zechengz, line 48: EARLIEST_TIMESTAMP?

torch_frame/typing.py
```diff
@@ -44,6 +44,9 @@ class NAStrategy(Enum):
...
+    MEDIAN_TIMESTAMP = 'median_timestamp'
```
zechengz, line 49: I think we can have MEDIAN (which is more general)?

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -685,3 +724,73 @@ def encode_forward(self, feat: TensorData) -> Tensor:
...
+        feat[:, :, TIME_TO_INDEX['DAYOFWEEK']] /= NUM_DAYS_PER_WEEK
+        feat[:, :, TIME_TO_INDEX['HOUR']] /= NUM_HOURS_PER_DAY
+        feat[:, :, TIME_TO_INDEX['MINUTE']] /= NUM_MINUTES_PER_HOUR
+        feat[:, :, TIME_TO_INDEX['SECOND']] /= NUM_SECONDS_PER_MINUTE
+        linear_component = feat[:, :, 0].unsqueeze(-1)
+        periodic_components = self.periodic(feat[:, :, 1:])
+        periodic_outputs = []
+        for c in self.mult_constants:
+            periodic_outputs.append(torch.sin(c * periodic_components))
+            periodic_outputs.append(torch.cos(c * periodic_components))
```
weihua916, line 793: Avoid for-loop.

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -685,3 +724,73 @@ def encode_forward(self, feat: TensorData) -> Tensor:
...
+        super().reset_parameters()
+        self.linear.reset_parameters()
+        self.periodic.reset_parameters()
+
+    def encode_forward(self, feat: Tensor) -> Tensor:
+        device = feat.device
+        feat[:, :,
+             TIME_TO_INDEX['YEAR']] = (feat[:, :, 0] - self.min_year.to(device)
+                                       ) / (self.max_year.to(device) -
+                                            self.min_year.to(device) + 1)
```
weihua916, line 781: Better formatting. Avoid min_year device transfer

weihua916: For year, can we not normalize?
just subtract the current year by the min year and apply Transformer's positional encoding (see [here](https://machinelearningmastery.com/a-gentle-introduction-to-positional-encoding-in-transformer-models-part-1/)). You may find some official implementation.

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -685,3 +724,73 @@ def encode_forward(self, feat: TensorData) -> Tensor:
...
+    """
+    supported_stypes = {stype.timestamp}
+
+    def __init__(
+        self,
+        out_channels: Optional[int] = None,
+        stats_list: Optional[List[Dict[StatType, Any]]] = None,
+        stype: Optional[stype] = None,
+        post_module: Optional[Module] = None,
+        na_strategy: Optional[NAStrategy] = None,
```
weihua916, line 746: Add `num_freqs` (number of frequencies) as an additional argument.
We will expand 7 dim input into `7*num_freqs` dim output and apply a linear function.

weihua916 (commented):
Left some initial comments :)

torch_frame/typing.py
yiweny: okay. Let me create a different pr to refactor.

test/nn/encoder/test_stypewise_encoder.py
```diff
@@ -96,17 +104,8 @@ def test_stypewise_feature_encoder(
...
-        "text_tokenized_1",
-        "text_tokenized_2",
-        "multicat_1",
-        "multicat_2",
-        "multicat_3",
-        "multicat_4",
+        "num_1", "num_2", "num_3", "cat_1", "cat_2", "text_embedded_1",
+        "text_embedded_2", "text_tokenized_1", "text_tokenized_2",
+        "multicat_1", "multicat_2", "multicat_3", "multicat_4", 'timestamp_0',
+        'timestamp_1', 'timestamp_2'
```
rusty1s, line 110:
```suggestion
        'timestamp_1', 'timestamp_2', 
```

torch_frame/data/mapper.py
rusty1s: I would prefer to keep the materialized data as integer, as they are int based, and we don't have to deal with floating point precision issues.

torch_frame/data/stats.py
```diff
@@ -139,9 +159,20 @@ def compute_col_stats(
...
         stats = {
-            stat_type: stat_type.compute(ser.dropna(), sep, time_format)
+            stat_type: stat_type.compute(ser.dropna(), sep)
             for stat_type in StatType.stats_for_stype(stype)
         }
 
     return stats
+
+
+def extract_time(t: datetime) -> List[np.float32]:
```
rusty1s, line 173:
```suggestion
def extract_time(t: datetime) -> List[int]:
```
Why `List[np.float32]`? You can also just return a `Tensor` here?

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -58,21 +87,18 @@ def __init__(
...
-            if (self.stype == stype.multicategorical
-                    and not self.na_strategy.is_multicategorical_strategy):
+            if ((self.stype == stype.numerical
+                 and not self.na_strategy.is_numerical_strategy)
+                    or (self.stype == stype.categorical
+                        and not self.na_strategy.is_categorical_strategy)
+                    or (self.stype == stype.multicategorical
+                        and not self.na_strategy.is_multicategorical_strategy)
+                    or (self.stype == stype.timestamp
+                        and not self.na_strategy.is_timestamp_strategy)):
```
rusty1s, line 97: What is this monster of a `if` clause? :D

torch_frame/stype.py
```diff
@@ -60,6 +60,13 @@ def use_dict_multi_nested_tensor(self) -> bool:
         """
         return self in [stype.text_tokenized]
 
+    @property
```
rusty1s, line 63: Should be a class property.

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -685,3 +725,71 @@ def encode_forward(self, feat: TensorData) -> Tensor:
...
+        super().reset_parameters()
+
+    def encode_forward(self, feat: Tensor) -> Tensor:
+        feat[:, :, TIME_TO_INDEX['YEAR']] = feat[:, :, 0] - self.min_year
+        feat[:, :, TIME_TO_INDEX['MONTH']] /= NUM_MONTHS_PER_YEAR
+        feat[:, :, TIME_TO_INDEX['DAY']] /= NUM_DAYS_PER_MONTH
+        feat[:, :, TIME_TO_INDEX['DAYOFWEEK']] /= NUM_DAYS_PER_WEEK
+        feat[:, :, TIME_TO_INDEX['HOUR']] /= NUM_HOURS_PER_DAY
+        feat[:, :, TIME_TO_INDEX['MINUTE']] /= NUM_MINUTES_PER_HOUR
+        feat[:, :, TIME_TO_INDEX['SECOND']] /= NUM_SECONDS_PER_MINUTE
```
rusty1s, line 786: Can we avoid the in-place modification here? I think it is not expected that input gets modified in-place.

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -685,3 +725,71 @@ def encode_forward(self, feat: TensorData) -> Tensor:
...
+        pe = torch.zeros((feat.size(0), feat.size(1),
+                          stype.timestamp.timestamp_num_cols, self.num_freqs))
+        position = self.position.repeat(feat.size(0), feat.size(1),
+                                        1).unsqueeze(-1)
+        pe[:, :, :, 0::2] = torch.sin(position * self.div_term)
+        pe[:, :, :,
+           1::2] = torch.cos(position * self.div_term[:self.num_freqs // 2])
```
rusty1s, line 793: Why not create as a buffer in advance?

torch_frame/nn/encoder/stype_encoder.py
yiweny: I don't think this is in-place modification. The tensor passed into this function will be cloned first. https://github.com/pyg-team/pytorch-frame/blob/master/torch_frame/nn/encoder/stype_encoder.py#L93

torch_frame/nn/encoder/stype_encoder.py
yiweny: I already created a buffer for the `div_term`. But position tensor shape depends on the batch_size and number of columns so it's hard to pre-determine the size and create a buffer in this case.

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -685,3 +730,86 @@ def encode_forward(self, feat: TensorData) -> Tensor:
...
+        pe[:, :, 1:,
+           1::2] = torch.cos(position * self.div_term[:self.num_freqs // 2])
+
+        year_position = feat[:, :, TIME_TO_INDEX['YEAR']].view(
+            feat.size(0), feat.size(1), 1, 1).repeat(1, 1, 1, self.num_freqs)
+        pe[:, :, 0:1,
+           0::2] = torch.sin(year_position[:, :, :, 0::2] * self.mult_term)
+        pe[:, :, 0:1, 1::2] = torch.cos(year_position[:, :, :, 1::2] *
+                                        self.mult_term[:self.num_freqs // 2])
+        feat = feat.unsqueeze(-1) + pe
```
weihua916, line 814: Let's add proper modules, test them thoroughly, and call them here.

weihua916 (commented):
I suggest us adding `nn.encoding` similar to https://pytorch-geometric.readthedocs.io/en/latest/_modules/torch_geometric/nn/encoding.html, test them thoroughly, and re-use them here.

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -685,3 +730,86 @@ def encode_forward(self, feat: TensorData) -> Tensor:
...
+        div_term = torch.exp(
+            torch.arange(0, self.num_freqs, 2).float() *
+            (-np.log(10000.0) / self.num_freqs))
+        self.register_buffer("div_term", div_term)
+        mult_term = 2 * torch.pi * torch.arange(
+            1, (self.num_freqs + 1) // 2 + 1).float()
+        self.register_buffer("mult_term", mult_term)
+        position = torch.arange(0, stype.timestamp.timestamp_num_cols - 1,
+                                dtype=torch.float)
+        self.register_buffer("position", position)
```
weihua916, line 782: Ideally, we just do the following:
```python
from torch_frame.nn.encoding import CyclicEncoding
from torch_frame.nn.encoding import PositionalEncoding

def init_module(self):
    ....
    self.cyclic_encoding = CyclicEncoding(num_freqs)
    self.positional_encoding = Positional(num_freqs)
 
def forward(self, feat):
    # normalize
    ....
    encoded_cyclic = self.cyclic_encoding(feat[:,:,:6])
    encoded_year = self.positional_encoding(feat[:,:,6])
    # stack above and apply torch.einsum or linear
    ....
```

torch_frame/nn/encoder/stype_encoder.py
rusty1s: Oh, any reason we do a copy there?

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -685,3 +730,86 @@ def encode_forward(self, feat: TensorData) -> Tensor:
...
+
+    def encode_forward(self, feat: Tensor) -> Tensor:
+        feat = feat.to(torch.float32)
+        feat[:, :, TIME_TO_INDEX['YEAR']] = feat[:, :, 0] - self.min_year
+        feat[:, :, TIME_TO_INDEX['MONTH']] /= NUM_MONTHS_PER_YEAR
+        feat[:, :, TIME_TO_INDEX['DAY']] /= NUM_DAYS_PER_MONTH
+        feat[:, :, TIME_TO_INDEX['DAYOFWEEK']] /= NUM_DAYS_PER_WEEK
+        feat[:, :, TIME_TO_INDEX['HOUR']] /= NUM_HOURS_PER_DAY
+        feat[:, :, TIME_TO_INDEX['MINUTE']] /= NUM_MINUTES_PER_HOUR
+        feat[:, :, TIME_TO_INDEX['SECOND']] /= NUM_SECONDS_PER_MINUTE
```
rusty1s, line 799: A small note on optimization: You can actually perform the division all at once, e.g.:
```python
max_values = torch.tensor([NUM_MONTHS_PER_YEAR, NUM_DAYS_PER_MONTH, ...])
feat = feat / max_values.view(1, -1)

File: torch_frame/nn/encoder/stype_encoder.py [hunk 28]
```diff
@@ -685,3 +730,86 @@ def encode_forward(self, feat: TensorData) -> Tensor:
...
+        ])
+        self.register_buffer("min_year", min_year)
+        div_term = torch.exp(
+            torch.arange(0, self.num_freqs, 2).float() *
+            (-np.log(10000.0) / self.num_freqs))
+        self.register_buffer("div_term", div_term)
+        mult_term = 2 * torch.pi * torch.arange(
+            1, (self.num_freqs + 1) // 2 + 1).float()
+        self.register_buffer("mult_term", mult_term)
+        position = torch.arange(0, stype.timestamp.timestamp_num_cols - 1,
```
rusty1s, line 780: I think you can replace `stype.timestamp.timestamp_num_cols` here just with `len(TIME_TO_INDEX)`. I think it is quite confusing to expose `timestamp_num_cols` as a property of `stype`.

torch_frame/stype.py
rusty1s: Resolved. Please see my other comment.

rusty1s (approved):
Looks mostly good, thanks :)

torch_frame/nn/encoder/stype_encoder.py
yiweny: This property is no longer necessary after we make the encoding modular

torch_frame/nn/encoder/stype_encoder.py
weihua916: Because we don't want to modify the original data via NaN imputation.

test/nn/encoder/test_stype_encoder.py
```diff
@@ -151,6 +152,33 @@ def test_multicategorical_feature_encoder(encoder_cls_kwargs):
...
+    ]
+
+    encoder = encoder_cls_kwargs[0](
+        8,
+        stats_list=stats_list,
+        stype=stype.timestamp,
+        **encoder_cls_kwargs[1],
+    )
+    feat_timestamp = tensor_frame.feat_dict[stype.timestamp]
+    x = encoder(feat_timestamp)
```
weihua916, line 178: nit: pass `col_names` just to follow the other test examples.

test/nn/encoder/test_stype_encoder.py
```diff
@@ -251,6 +279,45 @@ def test_multicategorical_feature_encoder_with_nan(encoder_cls_kwargs):
...
+        for col_name in tensor_frame.col_names_dict[stype.timestamp]
+    ]
+    encoder = encoder_cls_kwargs[0](
+        8,
+        stats_list=stats_list,
+        stype=stype.timestamp,
+        **encoder_cls_kwargs[1],
+    )
+    feat_timestamp = tensor_frame.feat_dict[stype.timestamp]
+    x = encoder(feat_timestamp)
```
weihua916, line 316: pass `col_names`

torch_frame/data/stats.py
```diff
@@ -145,9 +163,22 @@ def compute_col_stats(
...
+
+
+def extract_time(t: datetime) -> Tensor:
+    # subtracting one so that the smallest months and days can
+    # start from 0.
+    times = [
+        t.year, t.month - 1, t.day - 1,
+        t.weekday(), t.hour, t.minute, t.second
+    ]
+    return torch.tensor(times)
```
weihua916, line 184: Is there any way to call `TimestampTensorMapper` here, or use some shared function, similar to how we reuse https://github.com/pyg-team/pytorch-frame/blob/master/torch_frame/data/mapper.py#L140-L154 for multi-categorical?

weihua916: This way, we keep the ordering in the same unified place.

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -19,10 +20,32 @@
...
+NUM_SECONDS_PER_MINUTE = 60
+TIME_TO_INDEX = {
+    'YEAR': 0,
+    'MONTH': 1,
+    'DAY': 2,
+    'DAYOFWEEK': 3,
+    'HOUR': 4,
+    'MINUTE': 5,
+    'SECOND': 6
+}
```
weihua916, line 47: Can we derive this from `TimestampTensorMapper`? The idea is to keep this ordering in a single place.

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -182,6 +212,15 @@ def na_forward(self, feat: TensorData) -> TensorData:
...
                 fill_value = 0
+            elif self.na_strategy == NAStrategy.NEWEST_TIMESTAMP:
+                fill_value = self.stats_list[col][StatType.NEWEST_TIME].to(
+                    feat.device)
+            elif self.na_strategy == NAStrategy.OLDEST_TIMESTAMP:
+                fill_value = self.stats_list[col][StatType.OLDEST_TIME].to(
+                    feat.device)
+            elif self.na_strategy == NAStrategy.MEDIAN_TIMESTAMP:
+                fill_value = self.stats_list[col][StatType.MEDIAN_TIME].to(
+                    feat.device)
```
weihua916, line 223: Can you check why it blames these two are not covered in the test? I think you've covered it, but there may be some error?

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -744,3 +783,67 @@ def encode_forward(
         # -> [batch_size, num_cols, out_channels]
         x = x_lin + self.bias
         return x
+
+
+class TimestampEncoder(StypeEncoder):
+    r"""TimestampEncoder for timestamp stype. The encoder applies
+    different frequencies to the timestamps to calculate the
+    positional encoding and then adds the positional encoding to
+    the original :obj:`Tensor`. Year is encoded with
```
weihua916, line 792: Revise this as this is outdated.

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -744,3 +783,67 @@ def encode_forward(
...
+    :class:`torch_frame.nn.encoding.PositionalEncoding`. The other
+    features, including month, day, dayofweek, hour, minute and second,
+    are encoded using :class:`torch_frame.nn.encoding.CyclicEncoding`.
+    We then use a :class:`~torch.nn.Linear` layer to map back to
+    out_channels.
+
+    Args:
+        out_size (int): Output dimension of the positional and cyclic
+            encoding.
```
weihua916, line 801: nit: encodings

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -744,3 +783,67 @@ def encode_forward(
...
+            1, NUM_MONTHS_PER_YEAR, NUM_DAYS_PER_MONTH, NUM_DAYS_PER_WEEK,
+            NUM_HOURS_PER_DAY, NUM_MINUTES_PER_HOUR, NUM_SECONDS_PER_MINUTE
+        ])
+        feat[..., TIME_TO_INDEX['YEAR']] = feat[:, :, 0] - self.min_year
+        feat /= max_values.view(1, -1)
+        encodings = [
+            self.positional_encoding(feat[..., :1]),
+            self.cyclic_encoding(feat[..., 1:])
+        ]
+        out = torch.cat(encodings, dim=2).view(feat.size(0), feat.size(1), -1)
```
weihua916, line 848: nit: avoid using "-1", since it will error out when feat is emty tensor (can happen when neighbor smapling does not sample any node for this type)

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -744,3 +783,67 @@ def encode_forward(
...
+        max_values = torch.tensor([
+            1, NUM_MONTHS_PER_YEAR, NUM_DAYS_PER_MONTH, NUM_DAYS_PER_WEEK,
+            NUM_HOURS_PER_DAY, NUM_MINUTES_PER_HOUR, NUM_SECONDS_PER_MINUTE
+        ])
+        feat[..., TIME_TO_INDEX['YEAR']] = feat[:, :, 0] - self.min_year
+        feat /= max_values.view(1, -1)
+        encodings = [
+            self.positional_encoding(feat[..., :1]),
+            self.cyclic_encoding(feat[..., 1:])
+        ]
```
weihua916, line 847: nit: I prefer to split `feat` at the beggining and finally stack (you don't need view)

feat_year = feat[...,:1]
feat_rest = feat[..., 1:] / max_values.view(1, -1)
torch.stack([self.positional_encoding(feat_year), self.cyclic_encoding(feat_rest)], dim = 1)

torch_frame/nn/encoder/stype_encoder.py
yiweny: But isn't the original data already copied before the NaN imputation?

weihua916: I don't think so. Where did we copy?

yiweny: [here](https://github.com/pyg-team/pytorch-frame/blob/master/torch_frame/nn/encoder/stype_encoder.py#L115)

weihua916: Yes, I am aware of that. So the in-place imputation is safe, right?

torch_frame/nn/encoder/stype_encoder.py
yiweny: Yeah this is interesting.

yiweny: Turns out that I missed some parens. Added~

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -744,3 +775,68 @@ def encode_forward(
...
+        super().reset_parameters()
+        self.linear.reset_parameters()
+
+    def encode_forward(self, feat: Tensor,
+                       col_names: Optional[List[str]] = None) -> Tensor:
+        feat = feat.to(torch.float32)
+        max_values = torch.tensor([
+            NUM_MONTHS_PER_YEAR, NUM_DAYS_PER_MONTH, NUM_DAYS_PER_WEEK,
+            NUM_HOURS_PER_DAY, NUM_MINUTES_PER_HOUR, NUM_SECONDS_PER_MINUTE
+        ])
```
weihua916, line 831: sorry for being picky, but can this be also dependent on `TensorMapper`?

torch_frame/data/mapper.py
```diff
@@ -241,23 +242,45 @@ def backward(self, tensor: MultiNestedTensor) -> pd.Series:
...
+        'HOUR': 4,
+        'MINUTE': 5,
+        'SECOND': 6
+    }
+
     def __init__(self, format: str):
         super().__init__()
         self.format = format
 
+    @staticmethod
```
weihua916, line 259: can we unify extract time and forward? extract time can take ser as input.

torch_frame/data/stats.py
```diff
@@ -110,10 +119,18 @@ def compute(
...
             return [min(year_range), max(year_range)]
 
+        elif self == StatType.NEWEST_TIME:
+            return TimestampTensorMapper.extract_time(ser.iloc[-1])
+
+        elif self == StatType.OLDEST_TIME:
+            return TimestampTensorMapper.extract_time(ser.iloc[0])
+
+        elif self == StatType.MEDIAN_TIME:
+            return TimestampTensorMapper.extract_time(ser.iloc[len(ser) // 2])
```
weihua916, line 132: Here you can pass a series of length 1.

weihua916: rename extract_time to `to_tensor`

weihua916 (approved):
Thanks. I think it is super modular now!

Made some clear-up and fixes, especially on how we apply linear function.

PR #226, opened 2023-11-17
Update transformers text example

zechengz:
- Update transformers text example performance for some datasets
- Fix bug for saving file when there is `/` in the model name


yiweny (approved):
Thanks

weihua916 (approved):

PR #228, opened 2023-11-17
Text example fuse tokenizer and encoder to a single class

zechengz:
Fuse tokenizer and text encoder to avoid unnecessary errors such as tokenizer is not align with the text encoder


weihua916 (approved):
Thanks!

PR #229, opened 2023-11-18
Add `detach()` to `HashTextEmbedder`

weihua916:


zechengz (approved):

torch_frame/testing/text_embedder.py
```diff
@@ -29,4 +29,4 @@ def __call__(self, sentences: List[str]) -> Tensor:
         idx = torch.LongTensor(
             [hash(s) % self.num_hash_bins for s in sentences],
             device=self.device)
-        return self.embedding(idx)
+        return self.embedding(idx).detach()
```
rusty1s, line 32: Sorry for jumping in late. Can you clarify this change for me?

weihua916: Otherwise, `rtb` training will give us [this error](https://discuss.pytorch.org/t/runtimeerror-trying-to-backward-through-the-graph-a-second-time-but-the-buffers-have-already-been-freed-specify-retain-graph-true-when-calling-backward-the-first-time/6795) in the second training iteration.

rusty1s: Oh, this embedder here is not part of the network, right? Then this makes sense. Thanks!

weihua916: That's correct. It's pre-encoding of texts.

PR #234, opened 2023-11-21
Handle 0-row `TensorFrame` in `torch_frame.models`

weihua916:


torch_frame/nn/models/resnet.py
```diff
@@ -85,7 +86,7 @@ def forward(self, x: Tensor) -> Tensor:
         if self.shortcut is not None:
             x = self.shortcut(x)
 
-        out += x
+        out = out + x
```
zechengz, line 89: Just curious why not use +=

torch_frame/nn/models/resnet.py
```diff
@@ -194,7 +195,7 @@ def forward(self, tf: TensorFrame) -> Tensor:
         x, _ = self.encoder(tf)
 
         # Flattening the encoder output
-        x = x.view(x.size(0), -1)
+        x = x.view(x.size(0), math.prod(x.shape[1:]))
```
zechengz, line 198: Should this be same with -1?

zechengz (approved):

torch_frame/nn/models/resnet.py
weihua916: It gave me the error.

torch_frame/nn/models/resnet.py
weihua916: x is three dimensional, so it's not.

weihua916:
```bash
File "/home/weihua/code/kumo/venv/lib/python3.10/site-packages/torch/autograd/__init__.py", line 251, in backward
    Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
RuntimeError: one of the variables needed for gradient computation has been modified by an inplace operation: [torch.FloatTensor [0, 128]], which is output 0 of ReluBackward0, is at version 1; expected version 0 instead. Hint: enable anomaly detection to find the operation that failed to compute its gradient, with torch.autograd.set_detect_anomaly(True).
```

torch_frame/nn/models/resnet.py
weihua916:
```bash
File "/home/weihua/code/kumo/venv/lib/python3.10/site-packages/torch/autograd/__init__.py", line 251, in backward
    Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
RuntimeError: one of the variables needed for gradient computation has been modified by an inplace operation: [torch.FloatTensor [0, 128]], which is output 0 of ReluBackward0, is at version 1; expected version 0 instead. Hint: enable anomaly detection to find the operation that failed to compute its gradient, with torch.autograd.set_detect_anomaly(True).
```

torch_frame/nn/models/resnet.py
zechengz: I think they are same?
```python
>>> x = torch.arange(24).reshape(2, 3, 4)
>>> x.view(x.size(0), math.prod(x.shape[1:]))
tensor([[ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11],
        [12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23]])
>>> x.view(x.size(0), -1)
tensor([[ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11],
        [12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23]])
```

weihua916: Ah, this PR aims to resolve the corner case.
```
>>> torch.empty(0, 3, 10).view(0, -1)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
RuntimeError: cannot reshape tensor of 0 elements into shape [0, -1] because the unspecified dimension size -1 can be any value and is ambiguous
>>>
```

This can happen in GNN subgraph sampling, where we do not sample any nodes from a certain node type.

PR #236, opened 2023-11-22
Support `Metric` in `GBDT` + clean-up

weihua916:
so that we can use different `eval_metric` and `loss`, e.g., use MAE for regression instead of the default RMSE for regression, or use Accuracy for binary classification instead of the default ROC-AUC.

Also did some refactoring to clean-up the code.

### DIFF
--- a/torch_frame/gbdt/gbdt.py
+++ b/torch_frame/gbdt/gbdt.py
@@ -39,10 +39,10 @@ class GBDT:
         self._num_classes = num_classes
 
         # Set up metric
-        self._metric = DEFAULT_METRIC[task_type]
+        self.metric = DEFAULT_METRIC[task_type]
         if metric is not None:
             if metric.supports_task_type(task_type):
-                self._metric = metric
+                self.metric = metric
             else:
                 raise ValueError(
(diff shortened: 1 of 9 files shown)


torch_frame/gbdt/gbdt.py
```diff
@@ -81,41 +107,36 @@ def predict(self, tf_test: TensorFrame) -> Tensor:
...
-            return 'rmse'
-        elif self.task_type == TaskType.BINARY_CLASSIFICATION:
-            return 'rocauc'
-        elif self.task_type == TaskType.MULTICLASS_CLASSIFICATION:
-            return 'acc'
-        else:
-            raise ValueError(f"metric is not defined for {self.task_type}")
+    def metric(self) -> Metric:
+        r"""Evaluation metric to tune."""
+        return self._metric
```
rusty1s, line 112: Is there a need for this property? Why not just set `self.metric` in `__init__`?

rusty1s (approved):
Very clean, thanks!

torch_frame/gbdt/gbdt.py
weihua916: Good point. Deleted.

PR #239, opened 2023-11-24
Support for `stype.text_embedded` in `GBDT` models

rusty1s:


weihua916 (approved):
Thanks. Updated tests and added the support to catboost as well.

PR #243, opened 2023-11-28
Add `LinearMultiEmbeddingEncoder` for `embedding` stype

weihua916:
We can migrate `stype_encoder` of `text_embedded` into `LinearMultiEmbeddingEncoder`.
Then, we can rename `LinearMultiEmbeddingEncoder` into `LinearEmbeddingEncoder`, so that both `embedding` and `text_embedded` stypes use `LinearEmbeddingEncoder`.

### DIFF
--- a/torch_frame/nn/encoder/stype_encoder.py
+++ b/torch_frame/nn/encoder/stype_encoder.py
@@ -4,10 +4,18 @@ from typing import Any, Dict, List, Optional, Set
 
 import torch
 from torch import Tensor
-from torch.nn import Embedding, EmbeddingBag, ModuleList, Parameter, Sequential
+from torch.nn import (
+    Embedding,
+    EmbeddingBag,
+    ModuleList,
+    Parameter,
+    ParameterList,
+    Sequential,
+)
 from torch.nn.init import kaiming_uniform_
 
 from torch_frame import NAStrategy, stype
+from torch_frame.data.multi_embedding_tensor import MultiEmbeddingTensor
 from torch_frame.data.multi_nested_tensor import MultiNestedTensor
 from torch_frame.data.stats import StatType
 from torch_frame.nn.base import Module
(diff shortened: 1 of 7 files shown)


torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -562,16 +570,15 @@ def reset_parameters(self):
...
+    concatenates the output embeddings. Note that the implementation does this
+    for all :obj:`text_embedded` features in a batched manner.
+    NOTE: We will soon shift this to a new LinearMultiEmbeddingEncoder that
+    takes :class:`MultiEmbeddingTensor` object as input.
     """
-
-    # NOTE: We currently support text embeddings but in principle, this encoder
-    # can support any pre-encoded embeddings, including image/audio/graph
-    # embeddings.
+    # TODO: To be deprecated in favaor of LinearMultiEmbeddingEncoder
```
zechengz, line 581: Should we deprecate it? Is there any case that user want to use just one linear layer to transform the embeddings (for example user just use one kind of text encoder for all text columns)

zechengz (approved):
Thanks, in general LGTM.

torch_frame/nn/encoder/stype_encoder.py
weihua916: That's a good point. Personally, I think we can just deprecate this for simplicity. Otherwise, the type of `feat` needs to switch between `Tensor` and `MultiEmbeddingTensor`, which is confusing.

If single kind of text encoder is used, `LinearMultiEmbeddingEncoder` will essentially reduce to `LineariEmbeddingEncoder`.

weihua916: Also, I don't think we will have too many embedding-columns in practice, so for-loop over embedding column should not be too bad.

PR #246, opened 2023-11-28
Dictionary of `text_embedder_cfg` and use `MultiEmbeddingTensor` for `text_embedded`

### LINKED ISSUE #206: Allow different `text_embedder_cfg` for each column

zechengz:
Fixes #206

- Support dictionary of `text_embedder_cfg` and use MultiEmbeddingTensor for `text_embedded`
- Change `LinearEmbeddingEncoder`


test/datasets/test_fake_dataset.py
```diff
@@ -14,9 +14,26 @@
 @pytest.mark.parametrize('with_nan', [True, False])
 @pytest.mark.parametrize('tokenize_with_batch', [True, False])
 @pytest.mark.parametrize('text_batch_size', [None, 5])
-def test_fake_dataset(with_nan, tokenize_with_batch, text_batch_size):
+@pytest.mark.parametrize('use_different_embedder_cfg', [False, True])
```
weihua916, line 17: `use_per_col_embedder_cfg`

test/nn/encoder/test_stype_encoder.py
```diff
@@ -267,7 +266,7 @@ def test_text_embedded_encoder():
         for col_name in tensor_frame.col_names_dict[stype.text_embedded]
     ]
     # TODO: Use LinearMultiEmbeddingEncoder instead
-    encoder = LinearEmbeddingEncoder(
+    encoder = LinearMultiEmbeddingEncoder(
```
weihua916, line 269: nit: Remove TODO above.

torch_frame/data/dataset.py
```diff
@@ -111,12 +113,12 @@ class DataFrameToTensorFrameConverter:
             be used throughout all the multi-categorical columns. If a
             dictionary is given, we use a separator specified for each
             column. (default: :obj:`,`)
-        text_embedder_cfg
-            (:class:`torch_frame.config.TextEmbedderConfig`, optional):
-            A text embedder config specifying :obj:`text_embedder` that
-            maps sentences into :class:`torch.nn.Embeddings` and
-            :obj:`batch_size` that specifies the mini-batch size for
-            :obj:`text_embedder`. (default: :obj:`None`)
+        text_embedder_cfg (dict or TextEmbedderConfig, optional): A text
```
weihua916, line 116: `TextEmbedderConfigs, optional`

examples/llm_embedding.py
```diff
@@ -158,7 +158,7 @@ def __call__(self, sentences: List[str]) -> Tensor:
 stype_encoder_dict = {
     stype.categorical: EmbeddingEncoder(),
     stype.numerical: LinearEncoder(),
-    stype.text_embedded: LinearEmbeddingEncoder(),
+    stype.text_embedded: LinearMultiEmbeddingEncoder(),
```
weihua916, line 161: Can we rename all `LinearMultiEmbeddingEncoder` into `LinearEmbeddingEncoder` (given that we've deprecated the old `LinearEmbeddingEncoder`)?

torch_frame/data/dataset.py
```diff
@@ -525,14 +534,13 @@ def _update_col_stats(self):
...
-            for col_name in self._tensor_frame.col_names_dict[
-                    torch_frame.text_embedded]:
-                self._col_stats[col_name][StatType.EMB_DIM] = emb_dim
+            for i, col_name in enumerate(self._tensor_frame.col_names_dict[
+                    torch_frame.text_embedded]):
+                start = self._tensor_frame.feat_dict[
+                    torch_frame.text_embedded].offset[i]
+                end = self._tensor_frame.feat_dict[
+                    torch_frame.text_embedded].offset[i + 1]
+                self._col_stats[col_name][StatType.EMB_DIM] = int(end - start)
```
weihua916, line 543: nit:
outside of for-loop, we can do
```
offset = self._tensor_frame.feat_dict[torch_frame.text_embedded].offset
emb_dim_list = offset[1:] - offset[:-1]
```

torch_frame/data/dataset.py
```diff
@@ -143,7 +145,7 @@ def __init__(
         col_stats: Dict[str, Dict[StatType, Any]],
         target_col: Optional[str] = None,
         col_to_sep: Union[str, Dict[str, str]] = ",",
-        text_embedder_cfg: Optional[TextEmbedderConfig] = None,
+        text_embedder_cfg: Optional[TextEmbedderConfigs] = None,
```
weihua916, line 148: Shall we rename all `text_embedder_cfg` to `col_to_text_embedder_cfg`?

torch_frame/data/dataset.py
```diff
@@ -278,9 +284,12 @@ class Dataset(ABC):
             dictionary is given, we use a separator specified for each
             column. (default: :obj:`,`)
             (default: :obj:`,`)
-        text_embedder_cfg (TextEmbedderConfig, optional): A text embedder
-            configuration that specifies the text embedder to map text columns
-            into :pytorch:`PyTorch` embeddings. (default: :obj:`None`)
+        text_embedder_cfg (dict or TextEmbedderConfig, optional): A text
```
weihua916, line 287: `TextEmbedderConfigs, optional`

torch_frame/data/mapper.py
```diff
@@ -301,19 +301,30 @@ def forward(
...
             emb = self.text_embedder(ser_list[i:i + self.batch_size])
             emb_list.append(emb.to(device))
-        return torch.cat(emb_list, dim=0)
+        emb = torch.cat(emb_list, dim=0)
+        return MultiEmbeddingTensor(
+            num_rows=len(ser),
+            num_cols=1,
+            values=emb,
+            offset=torch.tensor([0, len(emb[0])]),
+        ).to(device)
```
weihua916, line 327: There are some code repetition in `MultiEmbeddingTensor` construction.
Can we do
```python
if self.batch_size is None:
   emb = ...
else:
   ...
   emb = torch.cat(emb_list, dim=0)
return MultiEmbeddingTensor(..., values = emb, ...)
```

torch_frame/data/dataset.py
```diff
@@ -205,9 +207,13 @@ def _get_mapper(self, col: str) -> TensorMapper:
...
+            if isinstance(self.text_embedder_cfg, dict):
+                text_embedder_cfg = self.text_embedder_cfg[col]
+            else:
+                text_embedder_cfg = self.text_embedder_cfg
             return TextEmbeddingTensorMapper(
-                self.text_embedder_cfg.text_embedder,
-                self.text_embedder_cfg.batch_size,
+                text_embedder_cfg.text_embedder,
+                text_embedder_cfg.batch_size,
             )
```
weihua916, line 217: Following the same pattern as `col_to_sep`, can we canonicalize beforehand, so that `self.text_embedder_cfgs` is always `dict`?
Here, we just do
```python
text_embedder_cfg = self.text_embedder_cfgs[col]
return TextEmbeddingTensorMapper(text_embedder_cfg.text_embedder, text_embedder_cfg.batch_size)
```

torch_frame/data/dataset.py
```diff
@@ -35,6 +35,8 @@
 )
 from torch_frame.utils.split import SPLIT_TO_NUM
 
+TextEmbedderConfigs = Union[Dict[str, TextEmbedderConfig], TextEmbedderConfig]
```
weihua916, line 38: Maybe, we don't need this additional typing to be more transparent.

weihua916 (commented):
Thanks! Looks good overall. Some suggestions:
- I think we can rename all `LinearMultiEmbeddingEncoder` into `LinearEmbeddingEncoder` (given that the old `LinearEmbeddingEncoder` is gone.)
- I think we can follow exactly the same pattern as `col_to_sep`, `col_to_time_format` here. So I suggest naming to `col_to_text_embedder_cfg`. It's a bit long name, but I think it's pretty informative. 
- Let's follow the same canonicalization pattern as in `col_to_sep`, `col_to_time_format`, so that `self.col_to_text_embedder_cfg` is always dictionary.

torch_frame/data/dataset.py
weihua916: Or follow https://github.com/pyg-team/pytorch-frame/pull/246#discussion_r1407325685

torch_frame/data/dataset.py
weihua916: or follow https://github.com/pyg-team/pytorch-frame/pull/246#discussion_r1407325685

weihua916 (approved):
Thanks! Made some fixes. LGTM.

PR #251, opened 2023-11-30
add cyclic encoding

yiweny:


torch_frame/nn/encoding/cyclic_encoding.py
```diff
@@ -0,0 +1,34 @@
...
+    Args:
+        out_size (int): The output dimension size.
+    """
+    def __init__(self, out_size: int):
+        super().__init__()
+        if out_size % 2 != 0:
+            raise ValueError(
+                f"out_size should be divisible by 2 (got {out_size}).")
+        self.out_size = out_size
+        mult_term = (torch.arange(0, self.out_size, 2) + 2) * math.pi
```
weihua916, line 26: for sine, I think k * pi is ok.
for cosine, we need 2 * k * pi

test/nn/encoding/test_cyclic_encoding.py
```diff
@@ -0,0 +1,24 @@
...
+        input_tensor = torch.rand(size)
+        cyclic_encoding = CyclicEncoding(out_size)
+        out_tensor = cyclic_encoding(input_tensor)
+        assert out_tensor.shape == input_tensor.shape + (out_size, )
+
+
+def test_cyclic_encoding_values():
+    out_size = 8
+    for size in [(10, ), (10, 4), (10, 5, 8)]:
+        input_tensor = torch.randint(0, 2, size=size)
```
weihua916, line 18: Can we generate fload input tensor?
Can we separately test 
```python
input_zeros_tensor = torch.zeros(size)
input_ones_tensor = torch.ones(size)
```
Gave exactly the same encoded output?

test/nn/encoding/test_cyclic_encoding.py
```diff
@@ -0,0 +1,24 @@
...
+        cyclic_encoding = CyclicEncoding(out_size)
+        out_tensor = cyclic_encoding(input_tensor)
+        assert torch.allclose(out_tensor[..., :4],
+                              torch.zeros_like(out_tensor[..., :4]), atol=1e-5)
+        assert torch.all(torch.abs(out_tensor[..., 4:] - 1) < 1e-5)
```
weihua916, line 23: These tests seem super hacky. I am not even sure why use `:4`

yiweny: because encoding size is 8. I can fix the code.

weihua916: I see. To me, just testing all ones and all zeros are encoded into the same values is enough.

weihua916 (approved):
Perfect, thanks!

PR #258, opened 2023-12-02
Fix GBDTs when `cat_features` is empty

xnuohz:


torch_frame/gbdt/tuned_lightgbm.py
```diff
@@ -18,7 +18,7 @@ class LightGBM(GBDT):
     def _to_lightgbm_input(
         self,
         tf: TensorFrame,
-    ) -> Tuple[np.ndarray, np.ndarray, List[int]]:
```
weihua916, line 21: I think the return type of the last element should be pure `List[int]`

torch_frame/gbdt/tuned_catboost.py
```diff
@@ -69,7 +69,8 @@ def _to_catboost_input(
             raise ValueError("The input TensorFrame object is empty.")
 
         df = pd.concat(dfs, axis=1)
-        cat_features = np.concatenate(cat_features, axis=0)
+        cat_features = np.concatenate(
+            cat_features, axis=0) if len(cat_features) else np.array([])
```
weihua916, line 73: Can you help fix the return type to be `np.ndarray` (no optional)?

weihua916 (approved):
Thanks! LGTM after return type is fixed

torch_frame/gbdt/tuned_lightgbm.py
```diff
@@ -30,7 +30,7 @@ def _to_lightgbm_input(
                 concatenating tensors of numerical and categorical features of
                 the input :class:`TensorFrame`.
             y (numpy.ndarray): Prediction target :obj:`numpy.ndarray`.
-            cat_features (List[int]): Array containing indexes of
+            cat_features (List[int], optional): Array containing indexes of
```
weihua916, line 33: same here. no optional

akihironitta (commented):
LGTM! Could we also have a test case to avoid regressions in the future?

xnuohz:
@weihua916 @akihironitta Fixed return types and improved test case.

weihua916 (approved):
LGTM!

PR #259, opened 2023-12-02
[Documentation] Typos & link fix

SimonPop:
Hi everyone!

I was reading through the introduction tutorial of the library and could spot minor typos. I decided to try and fix some of them.

At the same time, I updated the internal links to classes and methods when I could.

Here is the list of modifications proposed in this commit:
- Correcting minor typos.
- Updating internal links when the path is no longer correct (e.g. `torch_frame.TensorFrame` -> `torch_frame.data.TensorFrame`).
- Including the 's' for plural links into the pastil using a trick described in https://stackoverflow.com/questions/19403972/writing-terms-in-plural-without-the-redundancy (e.g. :class:`torch.Tensor`\ s -> :class:`Tensors<torch.Tensor>`).
- Providing links for existing classes previously unlinked (e.g. :obj:`Dataset` -> :obj:`torch_frame.data.Dataset`).

I don't know if tooling exists to maintain documentation when refactoring, but hopefully, this commit can refresh it slightly for now.

Do tell me if you think I have missed something that could be updated in this same commit, or if I did change something that was intended to be as it was.

Thank you for creating this library, I am very enthusiastic about it!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- Fixed documentation with updated links & corrected typos ([#259](https://github.com/pyg-team/pytorch-frame/pull/259))
 - Added `TimestampEncoder` ([#225](https://github.com/pyg-team/pytorch-frame/pull/225))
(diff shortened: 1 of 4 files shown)


weihua916 (approved):
Thank you so much! LGTM.

PR #263, opened 2023-12-03
Add `LightGBM` benchmark

xnuohz:
The result `{best_test_metric}±0.000 ({total_time}s)` is from the training log. Is this right?

### DIFF
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -25,12 +25,20 @@ scale=small
 # Specify the dataset idx from [0, 1, ...]
 idx=0
 
+# Specify the number of AutoML search trials
+num_trials=20
+
 # Specify the path to save the results
 result_path=results.pt
 
 # Run hyper-parameter tuning and training of the specified model on a specified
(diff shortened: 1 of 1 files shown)


weihua916:
Thank you! Looks great.

> The result {best_test_metric}±0.000 ({total_time}s) is from the training log. Is this right?

Yes, that's right. Below is my script
```
result = torch.load(os.path.join(dir, f'{model}.pt'))
mean = result['best_test_metric']
if 'best_test_metrics' in result:
    std = result['best_test_metrics'].std()
else:
    std = 0
time = result['total_time']
text = f"{mean:.3f}±{std:.3f} ({time:.0f}s)"
```

weihua916 (approved):
LGTM after addressing https://github.com/pyg-team/pytorch-frame/pull/263#issuecomment-1837634447

xnuohz:
@weihua916 @yiweny Done.

weihua916:
Thank you!

PR #264, opened 2023-12-03
Raise value error when numerical columns contain other dtypes

yiweny:
Hit the bug when running `california_house_price` dataset in `transformers_text.py`.
I think this is because some values are numbers, others are strings, e.g. `One guest bedroom` etc. So the series gets interpreted as `<U262` dtype.


torch_frame/data/stats.py
```diff
@@ -160,6 +160,12 @@ def compute_col_stats(
 ) -> Dict[StatType, Any]:
     if stype == torch_frame.numerical:
         ser = ser.mask(ser.isin([np.inf, -np.inf]), np.nan)
+        try:
+            np.isfinite(ser) | ser.isna()
+        except TypeError:
+            raise TypeError("Numerical series contains invalid entries."
+                            "Please make sure your numerical series "
+                            "contains only numerical values or nans.")
```
weihua916, line 168: Can't we use https://pandas.pydata.org/docs/reference/api/pandas.api.types.is_numeric_dtype.html?

weihua916 (approved):

akihironitta (approved):

PR #266, opened 2023-12-04
bug fix

yiweny:
Don't think they are needed because multicategorical support lists now


weihua916 (approved):
Thanks. Yes, we added the support at some point. LGTM once you've confirmed it works on those datasets.

PR #267, opened 2023-12-04
[WIP] Add ete example for timestamp

yiweny:


yiweny:
> Thanks! Can this be integrated into existing script that uses multi_modal benchmark?
> 
> Want to avoid too many example code / repetitive training scripts...

Don’t we want to convert text_embedded into embedding. I feel like that’s too mess to put all logic in one file. Actually, can we just remove text_embedded and use embedding instead

torch_frame/datasets/multimodal_text_benchmark.py
```diff
@@ -424,23 +426,19 @@ class MultimodalTextBenchmark(torch_frame.data.Dataset):
 
     def _pre_transform(self, df: pd.DataFrame,
                        target_col: str) -> pd.DataFrame:
-        if self.name == 'melbourne_airbnb':
-            df['host_verifications'] = df['host_verifications'].strip('[]')
-            df['amenities'] = df['amenities'].strip('[]')
```
weihua916, line 429: Why are we removing this?

torch_frame/datasets/multimodal_text_benchmark.py
```diff
@@ -20,6 +20,8 @@ class MultimodalTextBenchmark(torch_frame.data.Dataset):
         name (str): The name of the dataset to download.
         text_stype (torch_frame.stype): Text stype to use for text columns
             in the dataset. (default: :obj:`torch_frame.text_embedded`)
+        num_rows (int, optional): Number of rows to subsample.
+            (default: :obj:`None`)
```
weihua916, line 24: Why?

examples/tutorials/california_house_price.py
```diff
@@ -0,0 +1,197 @@
...
+                args.dataset)
+
+# Load Dataset
+dataset = MultimodalTextBenchmark(root=path, name=args.dataset, num_rows=5000, **kwargs)
+
+model_name = args.model.replace('/', '')
+filename = f"{model_name}_data.pt"
+dataset.materialize(path=osp.join(path, filename))
+dataset.tensor_frame.col_names_dict[stype.embedding] = dataset.tensor_frame.col_names_dict.pop(stype.text_embedded)
+dataset.tensor_frame.feat_dict[stype.embedding] = dataset.tensor_frame.feat_dict.pop(stype.text_embedded)
```
weihua916, line 88: These are not needed.

For the tutorial, we assume embeddings are given inside `df`, not something obtained in tensor frame.
I think we can just assume that embedding columns are given in the data frame (although in practice it's not).

weihua916: I am fine with adding some synthetic embedding columns just for illustration purposes in the tutorial html.

examples/tutorials/california_house_price.py
```diff
@@ -0,0 +1,197 @@
...
+            batch_size=5,
+        ),
+    }
+
+# Prepare datasets
+path = osp.join(osp.dirname(osp.realpath(__file__)), "..", "..", "data",
+                args.dataset)
+
+# Load Dataset
+dataset = MultimodalTextBenchmark(root=path, name=args.dataset, num_rows=5000, **kwargs)
```
weihua916, line 82: Hmm I don't really like `num_rows` argument --- it looks adhoc.

examples/tutorials/california_house_price.py
```diff
@@ -0,0 +1,197 @@
...
+val_loader = DataLoader(val_tensor_frame, batch_size=args.batch_size)
+test_loader = DataLoader(test_tensor_frame, batch_size=args.batch_size)
+
+stype_encoder_dict = {
+    stype.categorical: EmbeddingEncoder(),
+    stype.numerical: LinearEncoder(),
+    stype.embedding: LinearEmbeddingEncoder(),
+    stype.multicategorical: MultiCategoricalEmbeddingEncoder(),
+    stype.timestamp: TimestampEncoder(na_strategy=NAStrategy.MEDIAN_TIMESTAMP)
+}
```
weihua916, line 109: The biggest message in the tutorial should be this. We want to convey the message that this is all we need to handle more complex columns. Other parts are similar to https://pytorch-frame.readthedocs.io/en/latest/get_started/introduction.html so we can omit them in the tutorial.

torch_frame/datasets/multimodal_text_benchmark.py
yiweny: The full dataset was too large and take hours to run. But since we are moving to using a synthetic dataset we are all good.

yiweny:
Closing this because we plan on using sythetic dataset and no extra code. Will create a different pull request for the tutorial.

PR #268, opened 2023-12-04
allow for np.nan in multi categorical stype

yiweny:


weihua916 (approved):
Thanks! Can we update the test of `test_mapper.py` as well?

akihironitta (approved):

PR #271, opened 2023-12-06
Add documentation for different stypes

yiweny:


examples/stype_tutorial.py
yiweny, line 1: will delete this eventually before merge

torch_frame/data/dataset.py
```diff
@@ -360,20 +368,15 @@ def __init__(
...
-            col_to_sep,
-            [
-                col for col, stype in self.col_to_stype.items()
-                if stype == torch_frame.multicategorical
-            ],
-        )
+        self.col_to_sep = canonicalize_col_to_pattern(col_to_sep, [
+            col for col, stype in self.col_to_stype.items()
+            if stype == torch_frame.multicategorical
+        ], all_inclusive=False)
```
yiweny, line 374: If we have multiple multicategorical with a mixture of lists and string as raw data, then we were forced to specify a separator even for the list columns. This pr also changes this, so the specification doesn't need to be all inclusive.

CHANGELOG.md
```diff
@@ -8,7 +8,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
-- Fixed documentation with updated links & corrected typos ([#259](https://github.com/pyg-team/pytorch-frame/pull/259))
```
akihironitta, line 11: (nit: Deleted this entry as it doesn't impact users.)

docs/source/handling_advanced_stypes/handle_heterogeneous_stypes.rst
```diff
@@ -0,0 +1,242 @@
...
+    >>>
+        Numerical Categorical       Time                                  Multicategorical                                          Embedding
+    0         44      Type 2 2023-01-01              [Category D, Category A, Category B]  [0.2879910043632805, 0.38346222503494787, 0.74...
+    1         47      Type 2 2023-01-02  [Category C, Category A, Category B, Category D]  [0.0923738894608982, 0.3540466620838102, 0.551...
+    2         64      Type 2 2023-01-03                          [Category D, Category C]  [0.3209972413734975, 0.22126268518378278, 0.14...
+    3         67      Type 1 2023-01-04                          [Category C, Category A]  [0.2603409275874047, 0.5370225213757797, 0.447...
+    4         67      Type 2 2023-01-05                                      [Category A]  [0.46924917399024213, 0.8411401297855995, 0.90...
+
+
+Now let's load the :class:`pandas.DataFrame` into :class:`~torch_frame.data.Dataset` class.
```
akihironitta, line 61:
```suggestion
Now let's load the :class:`pandas.DataFrame` into :class:`torch_frame.data.Dataset` class so that we have :class:`~torch_frame.data.tensor_frame.TensorFrame` representation of the :class:`pandas.DataFrame`.
```

torch_frame/data/dataset.py
```diff
@@ -87,12 +90,13 @@ def canonicalize_col_to_pattern(col_to_pattern: Union[Optional[Any],
         for col in columns:
             col_to_pattern[col] = pattern
     else:
-        missing_cols = set(columns) - set(col_to_pattern.keys())
-        if len(missing_cols) > 0:
+        if all_inclusive and len(set(columns) -
```
akihironitta, line 93: It might worth adding a quick test case :)

akihironitta (commented):
🚀 (pushed small commiets but feel free to revert them)

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -788,7 +792,7 @@ def __init__(
         stats_list: list[dict[StatType, Any]] | None = None,
         stype: stype | None = None,
         post_module: Module | None = None,
-        na_strategy: NAStrategy | None = None,
+        na_strategy: NAStrategy | None = NAStrategy.MEDIAN_TIMESTAMP,
```
zechengz, line 795: Any reason why we make this na strategy as the default one?

yiweny: Because if no `NAStrategy` is specified, the encoder would fail on NaN timestamps. This happens quite often. So instead of raising an error I think it's better to provide a default strategy in the user experience perspective.

torch_frame/data/dataset.py
yiweny: Added.

weihua916 (approved):
Looks really great. thank you!

torch_frame/data/dataset.py
```diff
@@ -170,14 +172,18 @@ def __init__(
             # in-place sorting of col_names for each stype
             self._col_names_dict[stype].sort()
 
+        # Some multicategorical columns can be lists and thus
+        # do not need a separator.
         self.col_to_sep = canonicalize_col_to_pattern(
             col_to_sep,
-            self.col_names_dict.get(torch_frame.multicategorical, []),
+            self.col_names_dict.get(torch_frame.multicategorical),
+            all_inclusive=False,
```
weihua916, line 180: Can you clarify why we added `all_inclusive`? Seems to be set `False` for the main code.

torch_frame/data/dataset.py
```diff
@@ -212,8 +218,8 @@ def _get_mapper(self, col: str) -> TensorMapper:
             return CategoricalTensorMapper(index)
         elif stype == torch_frame.multicategorical:
             index, _ = self.col_stats[col][StatType.MULTI_COUNT]
-            return MultiCategoricalTensorMapper(index,
-                                                sep=self.col_to_sep[col])
+            return MultiCategoricalTensorMapper(
+                index, sep=self.col_to_sep.get(col, None))
```
weihua916, line 222: `sep = None` seems wrong.

yiweny: Discussed over slack. We agreed that it's weird for customers to specify a dummy string when the multicategorical column is a list. The `None` case refers to the case just mentioned. No strings splitting will occur in case of None.

PR #272, opened 2023-12-06
Use  `narrow(..., dim=1)` in favor of `[:, ...]`

akihironitta:
## Description of the change

Still to be benchmarked before merging this PR, but https://github.com/akihironitta/gist/blob/8ae6a01ecdb3471d2848c77aeeeb95b5a8288323/torch_contiguous/main.py suggests that:
```python
tensor.select(index, dim=1)
```
is faster than:
```python
tensor[:, index]
```
since `tensor[:, index]` first calls `aten::slice` followed by `aten::select` while the faster one only calls `aten::select`.

## Micro-benchmark result
```
[-------------- indexing -------------]
                                |  1024
1 threads: ----------------------------
      t[i]                      |  2.0
      t[i, :]                   |  4.6
      t[:, i]                   |  4.7
      t.select(dim=0, index=i)  |  2.6
      t.select(dim=1, index=i)  |  2.6

Times are in milliseconds (ms).
```

### DIFF
--- a/torch_frame/data/multi_embedding_tensor.py
+++ b/torch_frame/data/multi_embedding_tensor.py
@@ -121,10 +121,11 @@ class MultiEmbeddingTensor(_MultiTensor):
         offset = self.offset[start:start + length + 1] - self.offset[start]
         col_offset_start = self.offset[start]
         col_offset_end = self.offset[start + length]
+        col_index = slice(col_offset_start, col_offset_end)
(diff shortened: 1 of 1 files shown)


akihironitta:
Closing this as both `narrow(dim=1, start=start, length=length)` and [:, start:start+length]` call only `aten::slice`, and thus there's no performance difference.

PR #274, opened 2023-12-06
Update types to follow PEP 563, 585 and 604

akihironitta:
Updates existing type hints to make PyF compliant with:
- https://peps.python.org/pep-0585/
- https://peps.python.org/pep-0604/
- https://peps.python.org/pep-0563/

Most of the changes were automatically introduced by tools, but had to manually adjust some. It's compatible with Python 3.8 as we run Python 3.8 in our CI.

No need for a changelog entry as this doesn't impact users.

### DIFF
--- a/torch_frame/nn/models/excelformer.py
+++ b/torch_frame/nn/models/excelformer.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import Any
 
 import torch
 import torch.nn.functional as F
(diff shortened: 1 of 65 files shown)


yiweny (approved):

torch_frame/nn/models/excelformer.py
```diff
@@ -168,8 +170,8 @@ def forward(self, tf: TensorFrame) -> Tensor:
     def forward_mixup(
         self,
         tf: TensorFrame,
-        beta: Optional[float] = 0.5,
-    ) -> Union[Tensor, Tuple[Tensor, Tensor]]:
+        beta: float = 0.5,
+    ) -> Tensor | tuple[Tensor, Tensor]:
```
rusty1s, line 174: Note that this will break TorchScript support, but it probably doesn't matter.

PR #276, opened 2023-12-06
Type annotations for `MultiTensor`

akihironitta:
Add type hints to:
- `MultiEmbeddingTensor`
- `MultiNestedTensor`
- `_MultiTensor`

No need for a changelog entry as this PR doesn't directly impact users.


weihua916 (approved):

.pre-commit-config.yaml
```diff
@@ -62,4 +62,5 @@ repos:
     hooks:
       - id: mypy
         name: Check types
+        additional_dependencies: [torch==2.1.1]
```
akihironitta, line 65: This was necessary to resolve the discrepancy between pre-commit env (where `torch` wasn't installed) and the local dev env (where `torch` is installed).

torch_frame/data/multi_nested_tensor.py
```diff
@@ -164,7 +164,8 @@ def _col_narrow(self, start: int, length: int) -> MultiNestedTensor:
         accum = torch.cumsum(offset_mat_zero_start[:, -1], dim=0)
         offset_mat_zero_start[1:] += accum[:-1].view(-1, 1)
         num_cols = end - start
-        offset = torch.full((self.num_rows * num_cols + 1, ), accum[-1])
+        offset = torch.full((self.num_rows * num_cols + 1, ),
+                            cast(int, accum[-1]))
```
akihironitta, line 168: For reviewers, `typing.cast` is no-op at runtime and so introduces no impact on performance. See https://docs.python.org/3/library/typing.html#typing.cast as needed.

torch_frame/data/multi_tensor.py
```diff
@@ -38,12 +42,6 @@ def __setitem__(self, index: Any, values: Any) -> None:
             f"{self.__class__.__name__} object does not support setting "
             "values. It should be used for read-only.")
 
-    def __getitem__(
-        self,
-        index: Any,
-    ) -> _MultiTensor | Tensor:
-        raise NotImplementedError
```
akihironitta, line 45: mypy caught the duplicate definition of `__getitem__` I introduced in the past :) The real definition is below within this file.

torch_frame/data/multi_tensor.py
```diff
@@ -55,21 +53,18 @@ def __repr__(self) -> str:
     def shape(self) -> tuple[int, int, int]:
         return (self.num_rows, self.num_cols, -1)
 
-    @property
-    def ndim(self) -> int:
```
akihironitta, line 59: mypy thinks property is a `Callable` and thus caused an error at different places. Since `ndim` is constant across instances, I have changed it to a class variable.

akihironitta:
@weihua916 PTAL again if necessary :)

weihua916 (approved):
LGTM

PR #279, opened 2023-12-07
Drop mypy from pre-commit.ci and run mypy in a GHA workflow

akihironitta:
pre-commit.ci doesn't support having data more than 250MiB as seen in [this run](https://results.pre-commit.ci/run/github/677240069/1701948936.LN0QYu8HRtCf3o7UpKs6bQ):
```
Build of https://github.com/pre-commit/mirrors-mypy:torch==2.1.1@v1.7.1 for python@python3 exceeds tier max size 250MiB: 4.5GiB
```

This PR moves mypy from pre-commit.ci to GHA to unblock #276 and future PRs for type check in CI.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -1,0 +1,24 @@
+name: Linting
+
+on:  # yamllint disable-line rule:truthy
+  push:
+    branches:
+      - master
+  pull_request:
+
+jobs:
+  mypy:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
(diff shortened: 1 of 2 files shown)


weihua916 (approved):

.github/workflows/linting.yml
```diff
@@ -0,0 +1,24 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.8
+      - name: Install dependencies
+        run: |
+          pip install mypy
+          pip install torch --index-url https://download.pytorch.org/whl/cpu
```
rusty1s, line 22: I would assume you need to install the full `torch-frame` package here, as otherwise you are missing out on a lot of type checks of dependencies.

PR #281, opened 2023-12-08
Clean up `col_to_pattern` logic

weihua916:
- Remove redundancy, increase code sharing.
- Explicitly let `sep=None` handle `list` cell in multicategorical columns
- More type checking of canonicalized `col_to_pattern` to raise informative errors.

### DIFF
--- a/torch_frame/data/dataset.py
+++ b/torch_frame/data/dataset.py
@@ -191,7 +191,7 @@ class DataFrameToTensorFrameConverter:
             # in-place sorting of col_names for each stype
             self._col_names_dict[stype].sort()
 
-        # Assumed to be canonicalized already by dataset
+        # Assumed to be already canonicalized by Dataset
         self.col_to_sep = col_to_sep
         self.col_to_time_format = col_to_time_format
         self.col_to_text_embedder_cfg = col_to_text_embedder_cfg
--- a/torch_frame/data/dataset.py
+++ b/torch_frame/data/dataset.py
@@ -84,6 +84,7 @@ def requires_post_materialization(func):
 
 
 def canonicalize_col_to_pattern(
+    col_to_pattern_name: str,
     col_to_pattern: Any | dict[str, Any] | None,
     columns: list[str],
     requires_all_inclusive,
@@ -91,6 +92,8 @@ def canonicalize_col_to_pattern(
     r"""Canonicalize :obj:`col_to_pattern` into a dictionary format.
 
     Args:
+        col_to_pattern_name (str): The name of :obj:`col_to_xxx` to
+            canonicalize.
         col_to_pattern (Union[Any, Dict[str, Any]]): A dictionary or an object
             specifying the separator/pattern/configuration for the
             multi-categorical, timestamp or text columns. If an object is
@@ -121,8 +124,9 @@ def canonicalize_col_to_pattern(
         if len(missing_cols) > 0:
             if requires_all_inclusive:
                 raise ValueError(
-                    f"col_to_xxx requires all columns to be specified but the "
-                    f"following columns are missing: {list(missing_cols)}.")
+                    f"{col_to_pattern_name} requires all columns to be "
+                    f"specified but the following columns are missing from "
+                    f"the dictionary: {list(missing_cols)}.")
             else:
                 for col in missing_cols:
                     col_to_pattern[col] = None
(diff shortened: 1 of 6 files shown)


torch_frame/data/dataset.py
```diff
@@ -112,53 +140,44 @@ class DataFrameToTensorFrameConverter:
...
-            size for :obj:`text_tokenizer`. (default: :obj:`None`)
-        col_to_time_format (Union[str, Dict[str, str]], optional): A
-            dictionary or a string specifying the format for the timestamp
-            columns. See `strfttime documentation
+        col_to_sep (Dict[str, Optional[str]]): A dictionary specifying the
+            separator/delimiter for the multi-categorical columns.
+            (default: :obj:`{}`)
+        col_to_text_embedder_cfg (Dict[str, TextEmbedderConfig]):
+            A dictionary of configurations specifying :obj:`text_embedder` that
+            maps text columns into :class:`torch.nn.Embeddings` and
```
zechengz, line 148: I think it's not mapped to `:class:torch.nn.Embeddings`, let's change this.

torch_frame/data/dataset.py
```diff
@@ -91,13 +117,15 @@ def canonicalize_col_to_pattern(col_to_pattern: Any | dict[str, Any] | None,
...
-            raise ValueError(
-                "col_to_sep needs to specify separators for all "
-                "multi-categorical columns, but the separators for the "
-                "following columns are missing: "
-                f"{list(set(columns) - set(col_to_pattern.keys()))}.")
+        missing_cols = set(columns) - set(col_to_pattern.keys())
+        if len(missing_cols) > 0:
+            if requires_all_inclusive:
+                raise ValueError(
+                    f"col_to_xxx requires all columns to be specified but the "
```
zechengz, line 124: This `col_to_xxx` seems a little bit weird

torch_frame/data/dataset.py
```diff
@@ -112,53 +140,44 @@ class DataFrameToTensorFrameConverter:
...
         col_to_stype: dict[str, torch_frame.stype],
         col_stats: dict[str, dict[StatType, Any]],
         target_col: str | None = None,
-        col_to_sep: str | dict[str, str] = ",",
-        col_to_text_embedder_cfg: None |
-        (dict[str, TextEmbedderConfig] | TextEmbedderConfig) = None,
-        col_to_text_tokenizer_cfg: None |
-        (dict[str, TextTokenizerConfig] | TextTokenizerConfig) = None,
-        col_to_time_format: str | dict[str, str] | None = None,
+        col_to_sep: dict[str, str | None] = {},
```
zechengz, line 173: See this https://chat.openai.com/share/3506c8b0-4810-42f2-a23a-a9d4e3c38fe4

zechengz (approved):

yiweny (approved):

torch_frame/data/dataset.py
weihua916: I deliberately made so, since the input here is supposed to be canonicalized dictionary. By default, `col_to_pattern` should be an empty dictionary.

torch_frame/data/dataset.py
weihua916: Good point. will modify.

torch_frame/data/dataset.py
weihua916: Good catch. Rewrote as follows:
:obj:`text_embedder` that embeds texts into vectors

torch_frame/data/dataset.py
weihua916: Updated. Since empty dictionary will be never accessed, leaving them as `None` inside the class seems fine as well. Again, thanks for the good catch!

torch_frame/data/dataset.py
```diff
@@ -172,37 +200,11 @@ def __init__(
...
-                in self.col_names_dict) and (self.col_to_text_tokenizer_cfg
-                                             is None):
-            raise ValueError("`col_to_text_tokenizer_cfg` needs to be "
-                             "specified when stype.text_tokenized column "
-                             "exists.")
+        # Assumed to be already canonicalized by Dataset
+        self.col_to_sep = col_to_sep
+        self.col_to_time_format = col_to_time_format
+        self.col_to_text_embedder_cfg = col_to_text_embedder_cfg
+        self.col_to_text_tokenizer_cfg = col_to_text_tokenizer_cfg
```
yiweny, line 207: This will fail the `test_converter` test

PR #284, opened 2023-12-09
Allow `y=None` in `predict()` of `GBDT`

weihua916:
- At BP stage, we should not force users to supply `y` in TensorFrame.
- Also cache input to avoid redundant computation from tensor frame in every objective call.


yiweny (approved):
Thanks!

PR #285, opened 2023-12-10
Type annotations for `torch_frame.datasets.*`

akihironitta:


torch_frame/datasets/tabular_benchmark.py
```diff
@@ -386,14 +386,9 @@ class TabularBenchmark(torch_frame.data.Dataset):
     base_url = 'https://huggingface.co/datasets/inria-soda/tabular-benchmark/raw/main/'  # noqa
     # Dedicated URLs for large datasets
     base_url_large = 'https://huggingface.co/datasets/inria-soda/tabular-benchmark/resolve/main/'  # noqa
+    name_list = sorted(list(name_to_task_category.keys()))
 
-    @classmethod
-    @property
```
akihironitta, line 391: classmethod + property is deprecated in Python 3.11 https://github.com/python/cpython/pull/92379

torch_frame/datasets/data_frame_benchmark.py
```diff
@@ -539,7 +539,7 @@ class DataFrameBenchmark(torch_frame.data.Dataset):
           - Yandex(name='year')
           - 0.0%
     """
-    dataset_categorization_dict = {
+    dataset_categorization_dict: dict[str, dict[str, list[tuple]]] = {
```
yiweny, line 542: Just curious what are the benefits of adding this type hint. Would the type hint test break if someone adds a dataset that doesn't follow the type hint here?

torch_frame/datasets/mercari.py
```diff
@@ -49,7 +49,6 @@ def __init__(
         col_to_text_embedder_cfg: dict[str, TextEmbedderConfig]
         | TextEmbedderConfig | None = None,
     ) -> None:
-        self.dfs = dict()
```
yiweny, line 52: Thanks. Good catch.

torch_frame/datasets/yandex.py
```diff
@@ -78,6 +78,7 @@ def get_df_and_col_to_stype(
             # if the numpy_dict contains only categorical or numerical features
             features = (categorical_features if categorical_features
                         is not None else numerical_features)
+            assert features is not None
```
yiweny, line 81: Raise an error with a good error message rather than using assert here?

yiweny (approved):

torch_frame/datasets/data_frame_benchmark.py
akihironitta: Yes, it'll break the type check test, but I don't think there'll be a case when the type is different than this.

torch_frame/datasets/yandex.py
akihironitta: This error is for type narrowing, and I feel like this is more of an internal error that should never raised in user scripts.

PR #286, opened 2023-12-10
Remove implicit clone

### LINKED ISSUE #254: Remove implicit clone in stype encoders
Currently, we clone the `TensorData` in `StypeEncoders` to avoid in place modifications. But this is bad because it is wasteful. We want to remove that clone while avoiding inplace modification.

yiweny:
Fixes #254 

Summary:
1. Only clone when `na_strategy` is specified in stype encoder. Only clone the `values` of a `MultiTensor`.


test/nn/encoder/test_stype_encoder.py
```diff
@@ -22,13 +26,29 @@
...
     TimestampEncoder,
 )
-from torch_frame.testing.text_embedder import HashTextEmbedder
 from torch_frame.testing.text_tokenizer import (
     RandomTextModel,
     WhiteSpaceHashTokenizer,
 )
 
 
+def eq(td1: TensorData, td2: TensorData):
```
rusty1s, line 35: We can move this function to the package directly.

torch_frame/data/multi_nested_tensor.py
```diff
@@ -65,6 +65,13 @@ def validate(self):
         assert self.offset[-1] == len(self.values)
         assert len(self.offset) == self.num_rows * self.num_cols + 1
 
+    def __eq__(self, td: object) -> bool:
```
rusty1s, line 68: FYI: This is a bit different from how PyTorch tensors deal with equality. If we want to keep consistency, `__eq__` would need to return a mask, and equality would need to be checked via `tensor.equal()`/`torch.eq` and `tensor.allclose()`.

torch_frame/data/multi_tensor.py
```diff
@@ -325,6 +325,37 @@ def select(
     def _single_index_select(self, index: int, dim: int) -> _MultiTensor:
         raise NotImplementedError
 
+    def fillna(self, fill_value: Union[float, int]) -> None:
+        is_float = self.dtype.is_floating_point
+        if (is_float and isinstance(fill_value, int)
```
rusty1s, line 330: I don't see an issue filling with an integer for floating-point values. Do we need this check? Why not let PyTorch handle any errors downstream.

torch_frame/data/multi_tensor.py
```diff
@@ -325,6 +325,37 @@ def select(
     def _single_index_select(self, index: int, dim: int) -> _MultiTensor:
         raise NotImplementedError
 
+    def fillna(self, fill_value: Union[float, int]) -> None:
```
rusty1s, line 328: Would be great if we can allow column-level vectors here as well.

torch_frame/data/multi_tensor.py
```diff
@@ -325,6 +325,37 @@ def select(
...
+            fill_value: Any
+        """
+        self._fill_col(index, fill_value)
+
+    def _fill_col(
+        self,
+        index: int,
+        fill_value: Union[float, int],
+    ) -> None:
+        raise NotImplementedError
```
rusty1s, line 357: Any reason we need `_fill_col` and `fill_col`?

torch_frame/__init__.py
```diff
@@ -11,7 +11,7 @@
     embedding,
 )
 from .data import TensorFrame
-from .typing import TaskType, Metric, DataFrame, NAStrategy
+from .typing import TaskType, Metric, DataFrame, NAStrategy, TensorData
```
rusty1s, line 14: Where is this helpful?

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -201,6 +208,19 @@ def _single_index_select(
             )
         assert False, "Should not reach here."
 
+    def _fill_col(
+        self,
+        index: int,
+        fill_value: Union[float, int],
+    ) -> None:
```
rusty1s, line 215: Do you mean `fillna_col`?

torch_frame/data/multi_embedding_tensor.py
```diff
@@ -201,6 +208,19 @@ def _single_index_select(
...
+        value_index = slice(self.offset[index], self.offset[index + 1])
+        values = self.values[:, value_index]
+        if self.values.is_floating_point():
+            values[torch.isnan(values)] = fill_value
+        else:
+            values[values == -1] = fill_value
+        self.values[:, value_index] = values
```
rusty1s, line 222: Do we need this? Looks like everything before is zero-copy anyway.

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -37,6 +37,19 @@ def reset_parameters_soft(module: Module):
         module.reset_parameters()
 
 
+def get_nan_mask(tensor: Tensor):
```
rusty1s, line 40:
```suggestion
def get_nan_mask(tensor: Tensor) -> Tensor:
```

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -174,21 +182,11 @@ def na_forward(self, feat: TensorData) -> TensorData:
         """
         if self.na_strategy is None:
             return feat
-
+        if (isinstance(feat, _MultiTensor) and get_nan_mask(feat.values).any()
+                or isinstance(feat, Tensor) and get_nan_mask(feat).any()):
+            # Only clone the feat if there are NaNs in the data
+            feat = feat.clone()
```
rusty1s, line 188: It's probably more expensive to check this here than just doing the copy.

torch_frame/data/multi_tensor.py
rusty1s, line 328: I think it would help us to use PyTorch notation for in-place ops, e.g., in this case, this should be `fillna_`.

rusty1s, line 328:
```suggestion
    def fillna(self, fill_value: Union[float, int]) -> Self:
```

torch_frame/nn/encoder/stype_encoder.py
rusty1s, line 188: minor nit: The `clone` here would also clone offsets, while we would only need to clone the values.

torch_frame/data/multi_nested_tensor.py
yiweny: I feel like returning a mask would be inefficient here. Because then we need to convert the two `MultiTensor`s to their dense format. The current implementation is faster. Or do you mean return a tuple of mask, one for the `values` and one for the `offset`?

torch_frame/data/multi_tensor.py
yiweny: Removed.

torch_frame/data/multi_tensor.py
yiweny: Got it.

yiweny: I get self is not defined error if I change it to self.

torch_frame/__init__.py
yiweny: In `eq` function used by tests, but we can move that into package.

torch_frame/data/multi_embedding_tensor.py
yiweny: Thanks. What you say make sense!

torch_frame/nn/encoder/stype_encoder.py
yiweny: Changed the code to only clone the `values` tensor instead.

test/nn/encoder/test_stype_encoder.py
```diff
@@ -324,40 +333,6 @@ def test_timestamp_feature_encoder_with_nan(encoder_cls_kwargs):
     assert (~torch.isnan(x)).all()
 
 
-def test_text_embedded_encoder():
```
yiweny, line 327: This test is no longer necessary because `text_embedded` is converted to `embedding` during materialization.

torch_frame/data/multi_nested_tensor.py
```diff
@@ -275,6 +282,27 @@ def _single_index_select(self, index: int, dim: int) -> MultiNestedTensor:
         else:
             raise RuntimeError(f"Unsupported dim={dim} for index_select.")
 
+    def fillna_col(
+        self,
+        index: int,
+        fill_value: Union[int, float, Tensor],
+    ) -> None:
```
zechengz, line 289: Add docs that we inplace modify the tensor?

torch_frame/data/multi_tensor.py
```diff
@@ -325,6 +324,49 @@ def select(
     def _single_index_select(self, index: int, dim: int) -> _MultiTensor:
         raise NotImplementedError
 
+    def fillna_(self, fill_value: Union[int, float, Tensor]):
+        """Fill the :obj:`MultiTensor` with fill_value.
```
zechengz, line 328: Also, indicate this is an inplace operation?

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -37,6 +37,19 @@ def reset_parameters_soft(module: Module):
...
+def get_nan_mask(tensor: Tensor) -> Tensor:
+    r"""Obtains the NaN maks of the input :obj:`Tensor`.
+
+    Args:
+        tensor (Tensor): Input :obj:`Tensor`.
+    """
+    if tensor.is_floating_point():
+        nan_mask = torch.isnan(tensor)
+    else:
+        nan_mask = tensor == -1
```
zechengz, line 49: This means na mask rather than nan mask for categorical column right?

torch_frame/data/multi_nested_tensor.py
zechengz: I think we should try to keep consistent with PyTorch functions.

test/data/test_multi_embedding_tensor.py
```diff
@@ -99,6 +99,52 @@ def test_from_tensor_list():
     assert torch.allclose(met.offset, expected_offset)
     assert_equal(tensor_list, met)
 
+    met_with_nan = met.clone()
+    met_with_nan.values[0::2] = -1
```
weihua916, line 103: Does this mask the entire `met.values`?

test/data/test_multi_embedding_tensor.py
```diff
@@ -99,6 +99,52 @@ def test_from_tensor_list():
...
+    met_4 = met_with_nan
+    met_4.fillna_(torch.tensor([[100, 101, 102]]))
+    assert torch.all(
+        torch.eq(
+            met_4.values,
+            torch.tensor([
+                [100, 100, 100, 101, 101, 102],
+                [3, 4, 5, 8, 9, 11],
+            ])))
+    assert torch.all(torch.eq(met_4.offset, expected_offset))
```
weihua916, line 138: I don't understand these...

torch_frame/data/multi_nested_tensor.py
```diff
@@ -65,6 +65,13 @@ def validate(self):
         assert self.offset[-1] == len(self.values)
         assert len(self.offset) == self.num_rows * self.num_cols + 1
 
+    def __eq__(self, td: object) -> bool:
+        if not isinstance(td, MultiNestedTensor):
+            raise NotImplementedError
+        return bool(
+            torch.all(torch.eq(self.offset, td.offset))
+            and torch.all(torch.eq(self.values, td.values)))
```
weihua916, line 73: `torch.eq` --> `torch.allclose`

torch_frame/data/multi_nested_tensor.py
weihua916: +1. I don't get the behavior even from the test.

torch_frame/data/multi_nested_tensor.py
rusty1s: I am just saying you should rename this function to `eq` and `equal` :)

weihua916 (commented):
I think this PR can be much simpler. I am sorry but I don't understand why we are introducing fully-fledged fillna functionalities, while only a small fraction of it is really used for the main purpose of the PR.

Also, I personally had a different way of design. I am in favor of some redesign...

torch_frame/data/multi_tensor.py
```diff
@@ -325,6 +324,49 @@ def select(
...
+
+        Args:
+            index (int): A column index of the tensor to select.
+            fill_value (Union[int, float, Tensor]): Scalars or vector
+                tensor of size (num_cols,) or (1, num_cols) to replace
+                NaNs.
+        """
+        if isinstance(fill_value, Tensor):
+            if fill_value.dim(
+            ) != 0 and not fill_value.numel() == self.num_cols:
```
rusty1s, line 338: Do you mean `dim() != 1`?

rusty1s, line 338:
```suggestion
            if fill_value.dim(
            ) != 0 and fill_value.numel() != self.num_cols:
```

torch_frame/data/multi_tensor.py
```diff
@@ -325,6 +324,49 @@ def select(
...
+                raise ValueError(
+                    "fillna_ with Tensor of shape "
+                    f"{fill_value.shape} is not supported. "
+                    f"Please use a Tensor of shape ({self.num_cols},)"
+                    f" or (1, {self.num_cols}).")
+            if fill_value.dim() != 0:
+                if fill_value.dim() == 1:
+                    fill_value = fill_value.unsqueeze(0)
```
rusty1s, line 346: Not sure I get all this `dim` logic here. Why not just do `fill_value.view(-1)` at the start?

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -208,7 +220,28 @@ def na_forward(self, feat: TensorData) -> TensorData:
...
+                feat.fillna_col(col, fill_value)
+            else:
+                column_data = feat[:, col]
+                nan_mask = get_nan_mask(column_data)
+                if nan_mask.ndim == 2:
+                    nan_mask = nan_mask.any(dim=-1)
+                assert nan_mask.ndim == 1
+                assert len(nan_mask) == len(column_data)
+                if not nan_mask.any():
+                    continue
```
rusty1s, line 233:
```suggestion
```

test/nn/encoder/test_stype_encoder.py
```diff
@@ -44,10 +45,12 @@ def test_categorical_feature_encoder(encoder_cls_kwargs):
         stype=stype.categorical,
         **encoder_cls_kwargs[1],
     )
-    feat_cat = tensor_frame.feat_dict[stype.categorical]
+    feat_cat = tensor_frame.feat_dict[stype.categorical].clone()
```
weihua916, line 48: This does not really test anything since the `dataset` does not contain `nan` in the first place. Same for others.

yiweny: I am testing the in-place modification here. So `feat_cat` should be the same value has the `tensor_frame.feat_dict[stype.categorical]` after being passed to encoders

weihua916: Yes I understand. My point is that the original data does not contain any NaN, so there is likely no in-place modification at all. What's interesting to test is when the data contains NaN.

yiweny:
There are some confusions regarding this PR that we resolved offline. Basically, @weihua916 didn't realize that the pr is also trying to solve a bug regarding multicategorical nan masking #301 . I will make it more clear by separating out this pr into two. One to remove implicit clone and another to fix nan masking. This will target 0.2.1 release.

test/nn/encoder/test_stype_encoder.py
yiweny: Clarified offline that this is testing there there are no-inplace modifications without nans

test/nn/encoder/test_stype_encoder.py
```diff
@@ -421,11 +406,23 @@ def test_text_tokenized_encoder():
         stype=stype.text_tokenized,
         col_to_model_cfg=col_to_model_cfg,
     )
-    feat_text = tensor_frame.feat_dict[stype.text_tokenized]
+    feat_text = copy.deepcopy(tensor_frame.feat_dict[stype.text_tokenized])
```
rusty1s, line 409: Does `text_tokenized` do not use `MultiNestedTensor` yet?

rusty1s (approved):

test/nn/encoder/test_stype_encoder.py
yiweny: `text_tokenized` use a dictionary of `MultiEmbeddingTensor`, so `deepcopy` is necessary here.

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -174,20 +182,36 @@ def na_forward(self, feat: TensorData) -> TensorData:
...
-                nan_mask = torch.isnan(column_data)
+
+        # Since we are not changing the number of items in each column, it's
+        # faster to just clone the values, while reusing the same offset
+        # object.
+        if isinstance(feat, Tensor):
+            if get_na_mask(feat).any():
+                feat = feat.clone()
+            else:
+                return feat
```
weihua916, line 193: super nit: shall we assert when `feat` is of `dict` type? I know it's covered in `self.na_strategy is None`, but a reader may wonder why we don't cover `isinstance(feat, dict)` case

PR #302, opened 2023-12-15
Update docs

vid-koci:
A few minor changes for docs clarity. From the doc reading, there is currently no way to use both embedded and tokenised text in the same table? I.e. to embed one text column, but tokenise a different one?


akihironitta (approved):

zechengz (approved):

weihua916:
> From the doc reading, there is currently no way to use both embedded and tokenised text in the same table? I.e. to embed one text column, but tokenize a different one?

Good point. We certainly can. I can clarify that.

PR #311, opened 2023-12-20
Fix `TimestampEncoder` not using `CyclicEncoding` for cyclic features

akihironitta:
See title.


yiweny (approved):
Thanks. That's awkward.

weihua916:
Good catch!!

PR #318, opened 2023-12-22
missing norm_2 in excel_former

anas-rz:
I guess norm_2 was not being used in ExcelFormer.


weihua916:
Thanks for the catch! @yiweny Can you double-check the original implementation?

yiweny (approved):

yiweny:
Thanks for catching!

PR #320, opened 2023-12-25
Add type annotations to `torch_frame.utils.*`

akihironitta:
🎅🎁

### DIFF
--- a/torch_frame/utils/concat.py
+++ b/torch_frame/utils/concat.py
@@ -64,7 +64,7 @@ def _cat_tensor_data(td_list: list[TD], dim: int) -> TD:
     if len(td_list) == 0:
         raise ValueError("Cannot concatenate an empty list.")
     elif len(td_list) == 1:
-        return td_list[0]  # TensorData
+        return td_list[0]
     else:
         for td in td_list:
             if not isinstance(td, td_list[0].__class__):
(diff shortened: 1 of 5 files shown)


torch_frame/utils/concat.py
```diff
@@ -11,11 +12,27 @@
 from torch_frame.data.tensor_frame import TensorFrame
 from torch_frame.typing import TensorData
 
+T = TypeVar(
+    "T",
```
zechengz, line 16: What's the meaning of T here?

zechengz (approved):

torch_frame/utils/concat.py
akihironitta: I think `T` is a common name for a type variable although I'm open to renaming it :)

torch_frame/utils/concat.py
```diff
@@ -46,7 +64,7 @@ def _cat_tensor_data(td_list: list[TensorData], dim: int) -> TensorData:
     if len(td_list) == 0:
         raise ValueError("Cannot concatenate an empty list.")
     elif len(td_list) == 1:
-        return td_list[0]
+        return td_list[0]  # TensorData
```
akihironitta, line 67:
```suggestion
        return td_list[0]
```

torch_frame/utils/concat.py
zechengz: Oh, sorry I mean the `"T"` in `TypeVar`. Didn't know that we should have `"T"` in `TypeVar` to match with the `T`

PR #332, opened 2024-01-06
Fix `tab_transformer.py` evaluation bug

weihua916:
ROC AUC should not take argmax value

### DIFF
--- a/examples/tab_transformer.py
+++ b/examples/tab_transformer.py
@@ -1,9 +1,7 @@
 """Reported (reproduced, xgboost) results of of TabTransformer model based on
 Table 1 of original paper https://arxiv.org/abs/2012.06678.
 
-adult: 73.8 (76.05) batch_size: 128, lr: 0.0001, num_heads: 32, num_layers: 6
-bank-marketing: 93.4 (78.35, 81.00)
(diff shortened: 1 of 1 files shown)


examples/tab_transformer.py
```diff
@@ -1,9 +1,7 @@
 """Reported (reproduced, xgboost) results of of TabTransformer model based on
 Table 1 of original paper https://arxiv.org/abs/2012.06678.
 
-adult: 73.8 (76.05) batch_size: 128, lr: 0.0001, num_heads: 32, num_layers: 6
-bank-marketing: 93.4 (78.35, 81.00)
-dota2: 63.3 (58.28, 53.75)
+TODO: Update numbers.
```
weihua916, line 4: @yiweny Can you help run the script and update the numbers here? Maybe we can reproduce the numbers on the paper.

yiweny: Updated. Still cannot reproduce on dota2 dataset. Adult dataset we see an improved score(I am assuming their original experiments use round up score for roc-auc as well). But we do have a better score for bank marketing dataset.

yiweny (approved):

Issue #331, opened 2024-01-04
Select rows in `TensorFrame` with a bool `Tensor` mask


toenshoff:
I was wondering what the intended behavior of a `tensor_frame` is when passing a row-wise Boolean mask to `__getitem__`?
Based on pandas data frames and torch tensors a user may expect Boolean masks to work for a `tensor_frame` object.

Currently, the result depends on the specific `stypes` occurring in the frame. If no `_MultiTensor` columns are present the masking seems to work as expected. 
In the `_MultiTensor` classes row selection only works correctly for integer tensors but does not raise an error when given a Boolean mask. Instead, it yields an incorrect output. This may yield an invalid `tensor_frame`. Here is an example:

```python
from torch_frame.datasets import FakeDataset
from torch_frame import stype
import torch

mask = torch.tensor([True, False, True])

# Frame without _MultiTensor Columns
stypes = [stype.categorical, stype.numerical]
tf_1 = FakeDataset(num_rows=3, stypes=stypes).materialize().tensor_frame

print(f'Trying Boolean mask for stypes {stypes}')
tf_1[mask].validate()

# Frame with _MultiTensor Column
stypes = [stype.categorical, stype.multicategorical]
tf_2 = FakeDataset(num_rows=3, stypes=stypes).materialize().tensor_frame

print(f'Trying Boolean mask for stypes {stypes}')
tf_2[mask].validate()
```
We get the following output:
```
Trying Boolean mask for stypes [<stype.categorical: 'categorical'>, <stype.numerical: 'numerical'>]
Trying Boolean mask for stypes [<stype.categorical: 'categorical'>, <stype.multicategorical: 'multicategorical'>]
Traceback (most recent call last):
  File "/home/jan/git/torch/multinestedtest.py", line 21, in <module>
    tf_2[mask].validate()
  File "/home/jan/miniconda3/envs/torch/lib/python3.10/site-packages/torch_frame/data/tensor_frame.py", line 120, in validate
    raise ValueError(
ValueError: The length of elements in feat_dict are not aligned, got 3 but expected 2.
```

Should Boolean masks be supported? I think this would be convenient. If not, then some error should be thrown early when this is passed as input.

akihironitta:
Thanks for reporting this issue. I agree that indexing `TensorFrame` with a bool tensor is nice to have. For now, you should be able to work around it by:
```python
mask = torch.tensor([True, False, True])
﻿﻿mask = mask.nonzero().squeeze()
[...]

PR #334, opened 2024-01-08
support Boolean masks in _MultiTensor

toenshoff:
Added support for Boolean masks in `_MultiTensor` class based on issue https://github.com/pyg-team/pytorch-frame/issues/331


weihua916 (commented):
Thanks. Looks mostly good. 
Can we add column-wise boolean mask test and make sure that works as well?

weihua916 (approved):
Thanks!

PR #339, opened 2024-01-16
Support empty `TensorFrame({}, {})`

akihironitta:
This PR adds support for empty `TensorFrame` similarly to `pandas.DataFrame`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ venv/*
 *.out
 data/**
 catboost_info/
+.mypy_cache
--- a/.gitignore
+++ b/.gitignore
@@ -18,4 +18,3 @@ venv/*
 *.out
 data/**
 catboost_info/
-.mypy_cache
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -18,3 +18,4 @@ venv/*
 *.out
 data/**
 catboost_info/
+.mypy_cache
```
yiweny, line 21: Remove?

torch_frame/data/tensor_frame.py
```diff
@@ -84,9 +84,6 @@ def __init__(
 
     def validate(self) -> None:
         r"""Validates the :class:`TensorFrame` object."""
-        if len(self.feat_dict) == 0:
```
yiweny, line 87: Can we check instead if not only `feat_dict` or `col_names_dict` is empty?

.gitignore
akihironitta: Decoupled this to another PR #341. PTAL

akihironitta, line 21:
```suggestion
```

torch_frame/data/tensor_frame.py
akihironitta: It should already be covered by the check `self.feat_dict.keys() != self.col_names_dict.keys()` right below this line.

rusty1s (approved):

PR #345, opened 2024-01-30
Fix narrow issue in `MultiNestedTensor`

rusty1s:


torch_frame/data/multi_tensor.py
```diff
@@ -143,9 +143,8 @@ def _normalize_index(
         if isinstance(index, int):
             if index < 0:
                 index = index + max_entries
-            if is_slice_end and index < 0 or index > max_entries:
-                raise IndexError(f"{idx_name} index out of bounds!")
-            elif (not is_slice_end) and (index < 0 or index >= max_entries):
+            index = min(index, max_entries)
+            if index < 0:
```
weihua916, line 147: Is it normal to clamp in this way? why are we removing `is_slice_end`?

weihua916:
See here: https://github.com/pyg-team/pytorch-frame/blob/master/torch_frame/data/multi_tensor.py#L257-L261

Below, the slice end should allow 100, but normal indexing should not.
```python
>>> import numpy as np
>>> np.zeros(100)[:100]
array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
       0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
       0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
       0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
       0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
       0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.])
>>> np.zeros(100)[100]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
IndexError: index 100 is out of bounds for axis 0 with size 100
```

torch_frame/data/multi_tensor.py
rusty1s: I see. I only checked the narrow code path, which is broken due to these checks (see the test).

rusty1s:
@weihua916 PTAL.

akihironitta (approved):

rusty1s:
Merging this since I am blocked. @weihua916 I can add any feedback as a follow-up.

weihua916 (commented):
Looks good! thanks.

Issue #352, opened 2024-02-14
Improve current the split and support more split methods


XinweiHe:
Right now the split method for `Dateset` object is very basic. We should support more flexible splits

- [ ] Make the current split method less error prone through refactoring and adding more validation checks
- [ ] Support train/val split in addition to train/val/test split
- [ ] Support random split function for `Dateset` for better user experience
- [ ] Support temporal split, i.e. split based on a time column in the Dataset

PR #353, opened 2024-02-14
[Split] random split method improvement

XinweiHe:
For https://github.com/pyg-team/pytorch-frame/issues/352

### DIFF
--- a/test/utils/test_split.py
+++ b/test/utils/test_split.py
@@ -1,5 +1,7 @@
 import numpy as np
 
+from torch_frame.datasets import FakeDataset
+from torch_frame.typing import TrainingStage
 from torch_frame.utils.split import SPLIT_TO_NUM, generate_random_split
 
 
@@ -9,13 +11,26 @@ def test_generate_random_split():
     val_ratio = 0.1
     test_ratio = 0.1
 
-    split = generate_random_split(num_data, seed=42, train_ratio=train_ratio,
-                                  val_ratio=val_ratio)
-    assert (split == SPLIT_TO_NUM['train']).sum() == int(num_data *
-                                                         train_ratio)
-    assert (split == SPLIT_TO_NUM['val']).sum() == int(num_data * val_ratio)
-    assert (split == SPLIT_TO_NUM['test']).sum() == int(num_data * test_ratio)
+    split = generate_random_split(num_data, seed=42,
+                                  ratios=[train_ratio, val_ratio])
+    assert (split == SPLIT_TO_NUM.get(TrainingStage.TRAIN)).sum() == int(
+        num_data * train_ratio)
+    assert (split == SPLIT_TO_NUM.get(TrainingStage.VAL)).sum() == int(
+        num_data * val_ratio)
+    assert (split == SPLIT_TO_NUM.get(TrainingStage.TEST)).sum() == int(
+        num_data * test_ratio)
     assert np.allclose(
         split,
         np.array([0, 1, 0, 0, 0, 0, 0, 0, 2, 1, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0]),
     )
+
+
+def test_split_e2e_basic():
+    # TODO: Add several more test cases using @pytest.mark.parametrize
+    num_rows = 10
+    dataset = FakeDataset(num_rows=num_rows).materialize()
+    dataset.random_split([0.5, 0.2])
+    train_set, val_set, test_set = dataset.split()
+    train_set.num_rows, val_set.num_rows == (int(10 * 0.5), int(10 * 0.2))
+    if test_set is not None:
+        test_set.num_rows == 10 - int(10 * 0.5) - int(10 * 0.2)
--- a/torch_frame/data/dataset.py
+++ b/torch_frame/data/dataset.py
@@ -33,8 +33,9 @@ from torch_frame.typing import (
     IndexSelectType,
     TaskType,
     TensorData,
+    TrainingStage,
 )
-from torch_frame.utils.split import SPLIT_TO_NUM
+from torch_frame.utils.split import SPLIT_TO_NUM, generate_random_split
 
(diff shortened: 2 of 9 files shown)


test/utils/test_split.py
```diff
@@ -9,13 +11,26 @@ def test_generate_random_split():
...
     test_ratio = 0.1
 
-    split = generate_random_split(num_data, seed=42, train_ratio=train_ratio,
-                                  val_ratio=val_ratio)
-    assert (split == SPLIT_TO_NUM['train']).sum() == int(num_data *
-                                                         train_ratio)
-    assert (split == SPLIT_TO_NUM['val']).sum() == int(num_data * val_ratio)
-    assert (split == SPLIT_TO_NUM['test']).sum() == int(num_data * test_ratio)
+    split = generate_random_split(num_data, seed=42,
+                                  ratios=[train_ratio, val_ratio])
```
weihua916, line 15: why do we want this to be a list? I think the previous argument is clearer.

torch_frame/data/dataset.py
```diff
@@ -707,20 +708,25 @@ def get_split(self, split: str) -> Dataset:
...
+        val_set = self.get_split(TrainingStage.VAL)
+        test_set = self.get_split(TrainingStage.TEST)
+        if test_set.num_rows == 0:
+            test_set = None
+        return train_set, val_set, test_set
+
+    def random_split(self, ratios: list[float] | None = None):
+        split = generate_random_split(self.num_rows, ratios)
+        self.split_col = 'split'
+        self.df[self.split_col] = split
```
weihua916, line 729: I think it'd be nice to supply an argument to control the random seed of the random seed.

weihua916: Also, add doc-string so that people know how to use it.

torch_frame/typing.py
```diff
@@ -28,6 +28,12 @@ def supports_task_type(self, task_type: 'TaskType') -> bool:
         return self in task_type.supported_metrics
 
 
+class TrainingStage(Enum):
```
weihua916, line 31: Brief doc-string.

torch_frame/utils/split.py
```diff
@@ -1,30 +1,84 @@
...
+        raise ValueError("No more than three training splits is supported")
+    if len(ratio) < 1:
+        raise ValueError("At least two training splits are required")
+
+    for val in ratio:
+        if val < 0:
+            raise ValueError("'ratio' can not contain negative values")
+
+    if sum(ratio) - 1 > 1e-2:
+        raise ValueError("'ratio' exceeds more than 100% of the data")
```
weihua916, line 84: Could you clarify what's wrong with the old implementation? I think the old one is simpler.

test/utils/test_split.py
XinweiHe: This format is more aligned with split method in torch https://pytorch.org/docs/stable/_modules/torch/utils/data/dataset.html#random_split. And that to support split with only train and validation data with the new format, we are able to do `split(ratios=[0.8], ...)` while with the old format, we could only do `split(train_ratio=0.8, val_ratio=None, ...)` which is a bit weird, because this indicates a train/test split instead of a train/val split which is more common.

weihua916: Yeah but here we are actually assigning TRAIN to the first ratio, VAL to the second ratio, and so on. So I still think the previous one is better.

weihua916: torch's random split makes sense since it just splits the dataset. There is no train/val/test assignment happening under the hood.

torch_frame/data/dataset.py
```diff
@@ -695,7 +696,7 @@ def col_select(self, cols: ColumnSelectType) -> Dataset:
 
         return dataset
 
-    def get_split(self, split: str) -> Dataset:
+    def get_split(self, split: TrainingStage) -> Dataset:
```
weihua916, line 699: I think it's nicer to accept `str` as input here. It's sometimes troublesome to import `TrainingStage`.

weihua916:
Thanks for the effort. I am not sure the additional complexity is worthwhile though. I think the generalization to `ratios: list[float]` is not necessary here since each element has its role of train/val/test ratios.

torch_frame/data/dataset.py
XinweiHe: Maybe we can support both? i.e. `class TrainingStage(str, Enum)`  I feel like `TrainingStage` has its advantage in that no mistypes will occur that causes unexpected behavior, e.g. 'Train' vs 'train'. I feel like this is also the reason we support lots of Enum classes in the codebase

weihua916: Yes, we can support both.

weihua916: `Union[str, TrainingStage]`

test/utils/test_split.py
XinweiHe: In the previous format we always assume we want to have a test set. So
1. it does not have the flexibility to doing something like (train: 0.3, val: 0.1, test: 0.2)
2. when we do train: 0.8, val: 0.0. Then the inferred test ratio is 0.2. Which is a bit weird because in this case only train and val (test) split is well-defined. i.e. we only intend to split the data into 2 parts instead of 3 parts. But we end up with an additional validation set that is completely empty

We could also keep the previous format but support the case where val ratio could be `None` which indicates there is no validation set instead of an empty validation set if you have strong opinion here.

weihua916: I understand you. but I still want features based on needs.
I don't find these logics useful for now, and having new interface would break other parts. For instance, this PR does not modify `generate_random_split` used in `data_frame_benchmark.py` and elsewhere.

XinweiHe: > For instance, this PR does not modify generate_random_split used in data_frame_benchmark.py and elsewhere

This is still a draft now. I kind of want to sync with you on this first before changing other parts of the code

PR #358, opened 2024-02-20
Text benchmark dataset split

zechengz:
Make sure text benchmark datasets (from `MultimodalTextBenchmark`) split same with https://arxiv.org/abs/2111.02705

### DIFF
--- a/torch_frame/datasets/data_frame_text_benchmark.py
+++ b/torch_frame/datasets/data_frame_text_benchmark.py
@@ -432,6 +432,7 @@ class DataFrameTextBenchmark(torch_frame.data.Dataset):
 
         # Add split col
         df = dataset.df
+        # Random split for datasets not from MultimodalTextBenchmark:
         if class_name != 'MultimodalTextBenchmark':
             if SPLIT_COL in df.columns:
                 df.drop(columns=[SPLIT_COL], inplace=True)
@@ -442,9 +443,11 @@ class DataFrameTextBenchmark(torch_frame.data.Dataset):
             })
             df = pd.concat([df, split_df], axis=1)
         else:
-            # Manually split validation set from the train one:
             df = df.sort_values(by=[SPLIT_COL])
-            if len(df[SPLIT_COL].unique()) == 2:
+            num_unique_split = len(df[SPLIT_COL].unique())
+            assert num_unique_split > 1
+            # Manually split validation set from the train one:
+            if num_unique_split == 2:
                 ser = df[SPLIT_COL]
                 train_ser = ser[ser == SPLIT_TO_NUM['train']]
                 split_ser = generate_random_split(length=len(train_ser),
--- a/torch_frame/datasets/data_frame_text_benchmark.py
+++ b/torch_frame/datasets/data_frame_text_benchmark.py
@@ -432,22 +432,13 @@ class DataFrameTextBenchmark(torch_frame.data.Dataset):
 
         # Add split col
         df = dataset.df
-        # Random split for datasets not from MultimodalTextBenchmark:
-        if class_name != 'MultimodalTextBenchmark':
-            if SPLIT_COL in df.columns:
-                df.drop(columns=[SPLIT_COL], inplace=True)
-            split_df = pd.DataFrame({
-                SPLIT_COL:
-                generate_random_split(length=len(df), seed=split_random_state,
-                                      train_ratio=0.8, val_ratio=0.1)
-            })
-            df = pd.concat([df, split_df], axis=1)
-        else:
+        # Follow default split for datasets of MultimodalTextBenchmark:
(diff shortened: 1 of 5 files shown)


torch_frame/datasets/data_frame_text_benchmark.py
```diff
@@ -430,14 +432,31 @@ def __init__(
...
         df = dataset.df
-        if SPLIT_COL in df.columns:
-            df.drop(columns=[SPLIT_COL], inplace=True)
-        split_df = pd.DataFrame({
-            SPLIT_COL:
-            generate_random_split(length=len(df), seed=split_random_state,
-                                  train_ratio=0.8, val_ratio=0.1)
-        })
-        df = pd.concat([df, split_df], axis=1)
+        if class_name != 'MultimodalTextBenchmark':
```
vid-koci, line 435: Can you add a comment here on why this separate handling, for future reference?

torch_frame/datasets/data_frame_text_benchmark.py
```diff
@@ -430,14 +432,31 @@ def __init__(
...
+                split_ser = generate_random_split(length=len(train_ser),
+                                                  seed=split_random_state,
+                                                  train_ratio=0.9,
+                                                  val_ratio=0.1,
+                                                  include_test=False)
+                split_ser = np.concatenate([
+                    split_ser,
+                    np.full(len(df) - len(split_ser), SPLIT_TO_NUM['test'])
+                ])
+                df[SPLIT_COL] = split_ser
```
vid-koci, line 459: Can you add an else here? Maybe an exception or something, just to avoid silent failures if there are not exactly 2 unique split values?

torch_frame/utils/split.py
```diff
@@ -5,25 +5,35 @@
...
+                          include_test: bool = True) -> np.ndarray:
     r"""Generate a list of random split assignments of the specified length.
     The elements are either :obj:`0`, :obj:`1`, or :obj:`2`, representing
     train, val, test, respectively. Note that this function relies on the fact
     that numpy's shuffle is consistent across versions, which has been
     historically the case.
     """
-    assert train_ratio + val_ratio < 1
+    if include_test:
+        assert train_ratio + val_ratio < 1
```
vid-koci, line 17: If we're not including test, then train and val should add up to 1 (up to an eps), right? Or are we ignoring val_ratio completely ? in which case you can just check that `train_ratio < 1`.

vid-koci (approved):
Added a few comments, looks good

torch_frame/datasets/data_frame_text_benchmark.py
```diff
@@ -430,14 +432,34 @@ def __init__(
...
+                df.drop(columns=[SPLIT_COL], inplace=True)
+            split_df = pd.DataFrame({
+                SPLIT_COL:
+                generate_random_split(length=len(df), seed=split_random_state,
+                                      train_ratio=0.8, val_ratio=0.1)
+            })
+            df = pd.concat([df, split_df], axis=1)
+        else:
+            df = df.sort_values(by=[SPLIT_COL])
+            num_unique_split = len(df[SPLIT_COL].unique())
```
weihua916, line 447: `df[SPLIT_COL].nunique`

torch_frame/datasets/data_frame_text_benchmark.py
```diff
@@ -430,14 +432,34 @@ def __init__(
...
-        if SPLIT_COL in df.columns:
-            df.drop(columns=[SPLIT_COL], inplace=True)
-        split_df = pd.DataFrame({
-            SPLIT_COL:
-            generate_random_split(length=len(df), seed=split_random_state,
-                                  train_ratio=0.8, val_ratio=0.1)
-        })
-        df = pd.concat([df, split_df], axis=1)
+        # Random split for datasets not from MultimodalTextBenchmark:
+        if class_name != 'MultimodalTextBenchmark':
```
weihua916, line 436: nit: Can you flit the logic? if `class_name == 'MultimodalTextBenchmark'`, else

torch_frame/utils/split.py
```diff
@@ -5,25 +5,37 @@
...
+    else:
+        assert train_ratio < 1
```
weihua916, line 19: Should be `assert train_ratio + val_ratio == 1`?

weihua916 (approved):
Thanks!

torch_frame/utils/split.py
zechengz: I was afraid that there maybe some numerical error. But will use math.isclose to check that.

weihua916: I don't think you need to worry about that. People will pass something like `0.9 + 0.1`

zechengz: ok

PR #360, opened 2024-02-20
Add benchmarking suite for encoders

DamianSzwichtenberg:
Preliminary benchmarks of [RelBench](https://github.com/snap-stanford/relbench/blob/main/examples/gnn.py) showed that a lot of execution time is spent on encoders. This PR introduces encoders benchmarking suite that can server as a tool for performance inspection.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


weihua916 (approved):
Thank you for the effort! What's the finding here? Which part of encoder is the bottleneck?

DamianSzwichtenberg:
> Thank you for the effort! What's the finding here? Which part of encoder is the bottleneck?

It heavily depends on the settings, encoder type, number of columns, and so on. The first optimization will try to fix looping over columns in `encode_forward`.

PR #361, opened 2024-02-20
Add support for a few common LLMs in the text benchmark script

vid-koci:
Adds support for RoBERTa, Electra, and Deberta finetuning. Picks the batch size depending on the model.

### DIFF
--- a/benchmark/data_frame_text_benchmark.py
+++ b/benchmark/data_frame_text_benchmark.py
@@ -7,7 +7,7 @@ import os.path as osp
 from typing import Any
 
 import torch
-from peft import LoraConfig, TaskType as peft_TaskType, get_peft_model
+from peft import LoraConfig, TaskType as peftTaskType, get_peft_model
 from torch import Tensor
 from torch.nn import BCEWithLogitsLoss, CrossEntropyLoss, Module, MSELoss
 from torch.optim.lr_scheduler import ExponentialLR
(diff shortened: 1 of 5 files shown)


benchmark/data_frame_text_benchmark.py
```diff
@@ -73,6 +75,9 @@
     default="sentence-transformers/all-distilroberta-v1",
     choices=[
         "distilbert-base-uncased",
+        "roberta-large",
+        "microsoft/deberta-v3-large",
```
zechengz, line 79: For now maybe we can just follow the paper https://arxiv.org/pdf/2111.02705.pdf just include roberta and electra

benchmark/data_frame_text_benchmark.py
```diff
@@ -131,11 +153,10 @@ def __init__(self, model: str):
         elif model == "sentence-transformers/all-distilroberta-v1":
             target_modules = ["intermediate.dense"]
         else:
-            raise ValueError(f"Model {model} is not specified for "
-                             f"LoRA finetuning.")
+            target_modules = "all-linear"
```
zechengz, line 156: Do we have experiments on finetuning which (sub)layers can achieve better performance?

zechengz (approved):

weihua916 (approved):
Thanks!

benchmark/data_frame_text_benchmark.py
vid-koci: Not yet, hence the default.

benchmark/data_frame_text_benchmark.py
vid-koci: I've added DeBERTa, but you're right, I don't think we'll be running experiments with it unless necessary or unless we have plently of extra time, which usually doesn't happen.

PR #363, opened 2024-02-23
Add roberta pooling layer to the trainable arguments

vid-koci:
Roberta randomly initialises the top pooling layer when loaded from transformers. Ideally, this layer should be trained in full, and not just with LoRa, to avoid random noise.

### DIFF
--- a/benchmark/data_frame_text_benchmark.py
+++ b/benchmark/data_frame_text_benchmark.py
@@ -152,6 +152,10 @@ class TextToEmbeddingFinetune(torch.nn.Module):
             target_modules = ["ffn.lin1"]
         elif model == "sentence-transformers/all-distilroberta-v1":
             target_modules = ["intermediate.dense"]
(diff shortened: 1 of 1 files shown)


benchmark/data_frame_text_benchmark.py
```diff
@@ -166,6 +170,10 @@ def __init__(self, model: str):
             target_modules=target_modules,
         )
         self.model = get_peft_model(self.model, peft_config)
+        if model == 'roberta-large':
```
zechengz, line 173: Actually we don't use the pooling results not, instead we average over the last hidden states. Actually we can try to use the pooling layer results and also finetune the pooler.

zechengz (approved):
Left some comments, in general LGTM.

vid-koci:
Let me, in that case, not merge this PR and we can add this if we actually decide to use the pooler. To avoid clutter in the code.

PR #364, opened 2024-02-26
Improve performance of Encoder's `na_forward` method

DamianSzwichtenberg:
This PR introduces small improvements that resulted in an average 1.5x speedup in `na_forward` execution time.
`get_na_mask` is a heavy operation, whose result may be cached at the pre-condition checking step.
Additionally, when we deal with 2D `feat`, we can use `torch.where` to efficiently fill `NaN`s across all columns.

### DIFF
--- a/torch_frame/nn/encoder/stype_encoder.py
+++ b/torch_frame/nn/encoder/stype_encoder.py
@@ -171,13 +171,14 @@ class StypeEncoder(Module, ABC):
         return out
 
     def na_forward(self, feat: TensorData) -> TensorData:
-        r"""Replace NaN values in input :obj:`Tensor` given :obj:`na_strategy`.
+        r"""Replace NaN values in input :obj:`TensorData` given
+        :obj:`na_strategy`.
 
         Args:
-            feat (Tensor): Input :obj:`Tensor`.
+            feat (TensorData): Input :obj:`TensorData`.
 
         Returns:
-            torch.Tensor: Output :obj:`Tensor` with NaNs replaced given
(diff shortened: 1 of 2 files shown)


torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -231,16 +234,22 @@ def na_forward(self, feat: TensorData) -> TensorData:
...
-                column_data[na_mask] = fill_value
+        else:
+            if na_mask.ndim == 3:
+                # when feat is 3D, it is faster to iterate over columns
+                for col, fill_value in enumerate(fill_values):
+                    col_data = feat[:, col]
+                    col_na_mask = na_mask[:, col].any(dim=-1)
+                    col_data[col_na_mask] = fill_value
+            else:  # na_mask.ndim == 2
+                fill_values = torch.tensor(fill_values, device=feat.device)
```
weihua916, line 250: `torch.stack(fill_values)`?

weihua916 (approved):
Thanks!

torch_frame/nn/encoder/stype_encoder.py
DamianSzwichtenberg: `torch.stack` would be correct for `na_mask.ndim == 3`. Here it would generate an error, because `fill_values` is a list of ints/floats, not a list of tensors. 😉

weihua916: Ah I see. Thanks for clarifying

PR #366, opened 2024-02-27
Avoid for-loop in `EmbeddingEncoder`

weihua916:


torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -284,39 +284,45 @@ def __init__(
...
-        # [batch_size, num_cols, hidden_channels]
-        x = torch.stack(xs, dim=1)
-        return x
+        # Get NaN mask
+        na_mask = feat < 0
+        # Increment the index by one not to conflict with the padding idx
+        # Also add offset for each column to avoid embedding conflict
+        feat = feat + self.offset + 1
+        # Use 0th index for NaN
+        feat[na_mask] = 0
```
rusty1s, line 323: Is it intended to use the same `N/A` value across all columns?

weihua916: They are mapped to all-zero vector.

weihua916: `padding_idx=0` in `Embedding`

rusty1s: Got it :)

rusty1s (approved):

PR #371, opened 2024-03-06
Add gradient accumulation and gradient clipping to text benchmark script

zechengz:
Add gradient accumulation and gradient clipping to text benchmark script
TODO: add warmup


vid-koci (approved):

zechengz:
Close this as it's not necessary.

PR #376, opened 2024-03-11
Handle label imbalance in binary classification tasks on text benchmark

vid-koci:
Labels in the text benchmarks are imbalanced and weighting the positive labels improves performance.
Experiments done on `fake` dataset (5% positive labels)  with `text_embedded` and `RoBERTa` encodings:

- `ResNet` result changes 91.1% -> 93.4% 
- `FTTransformer` result remains unchanged
- `Trompt` result changes 95.2% -> 95.8%

The differences were even more stark with distilled roberta, but we aren't reporting those anywhere so I didn't note them down.

More results are pending


weihua916 (approved):
Thanks!

PR #383, opened 2024-04-02
Add citation

weihua916:


zechengz (approved):
Should we also update the https://github.com/pyg-team/pytorch-frame/blob/master/CITATION.cff ?

akihironitta (approved):

PR #390, opened 2024-04-20
Add device support for XGB tuning

adobles96:
Add `device` kwarg to `XGBoost._tune` and `XGBoost.objective` to enable GPU training.


xnuohz:
Maybe it is better to add device in GBDT class, since all these models support GPU training.

weihua916:
Agree with @xnuohz . Do you want to take it over (if possible)?

adobles96:
Sure I can do that. I'll make it draft PR while I look into the other library training APIs as I'm not too familiar with those.

weihua916:
If @xnuohz has the bandwidth, could you help with it? @adobles96  Is that okay?

xnuohz:
Sure. Should I open a new PR for this?

weihua916:
@xnuohz thank you. Yes you can open a new PR.

PR #391, opened 2024-04-22
Update the key implementation and corresponding compatibility for ExcelFormer

jyansir:
Hello, I am an author and the code contributor of ExcelFormer.

I have adjusted the implementation of ExcelFormer to its real version, the related supporting and dependent scripts are updated accordingly. All scripts have passed the tests under PyTorch 2.2.0, 2.1.0 and 1.1.0 (Python 3.10) with the same coding style as the project.

The update descriptions are as follows:

**implementation update for ExcelFormer**: `torch_frame/nn/models/excelformer.py`.
- In the paper, the specifically designed two novel mixup methods, `FEAT-MIX` and `HIDDEN-MIX`, have been added. Their operations only consist of PyTorch matrix computation on encoded numerical features, thus I have integrated them into `forward` function to achieve further acceleration by `torch.compile`. For ordinary mixup on input numerical features, I follow the previous codes that include it in `forward_mixup` function, since the `Dict` operations cannot be compiled by `torch.compile`. I design to apply `FEAT-MIX` or `HIDDEN-MIX` through `forward_mixup` function thus it follows the previous calling style.

**compatibility adjustment for ExcelFormer**: `examples/excelformer.py`, `test/nn/models/test_excelformer.py`, `benchmark/data_frame_benchmark.py`, `torch_frame/transforms/mutual_information_sort.py`.
- `torch_frame/transforms/mutual_information_sort.py` is for feature sorting in ExcelFormer. In `FEAT-MIX`, the ExcelFormer requires the mutual information scores to calculate the mixup weights, thus I add several lines to assign the `mi_scores` as an attribute to `TensorFrame` object like setting the lazy attributes if fit and transform with `MutualInformationSort` class. It is also can be used as a meta feature for other models in the future when calling `MutualInformationSort` class with no impact on other behaviors.

**loss function for soft multi-class labels**: `torch_frame/nn/utils/loss.py`.
- Since in the paper ExcelFormer works with soft mixup weights, when encountering multi-class tasks, there will be mixed soft targets, which are not supported by `F.cross_entropy` or `CrossEntropyLoss`, I add a soft version. This file can be used to include other loss functions like Focal Loss in the future.


examples/excelformer.py
```diff
@@ -99,9 +102,11 @@ def train(epoch: int) -> float:
 
     for tf in tqdm(train_loader, desc=f'Epoch: {epoch}'):
         tf = tf.to(device)
+        # Train with FEAT-MIX or HIDDEN-MIX
         pred_mixedup, y_mixedup = model.forward_mixup(tf)
         if is_classification:
-            loss = F.cross_entropy(pred_mixedup, y_mixedup)
+            # Softly mixed one-hot labels
+            loss = cross_entropy_with_logits(pred_mixedup, y_mixedup)
```
weihua916, line 109: I believe that's what we are doing here? `y_mixedup` is already a soft label.

examples/excelformer.py
```diff
@@ -20,17 +20,19 @@
...
 from torch_frame.datasets.yandex import Yandex
 from torch_frame.nn import ExcelFormer
+from torch_frame.nn.utils.loss import cross_entropy_with_logits
 from torch_frame.transforms import CatToNumTransform, MutualInformationSort
 
 parser = argparse.ArgumentParser()
-parser.add_argument('--dataset', type=str, default='higgs_small')
+parser.add_argument('--dataset', type=str, default='california_housing')
+parser.add_argument('--mixup', type=str, default='feature',
+                    choices=['none', 'ordinary', 'feature', 'hidden'])
```
weihua916, line 29: What is ordinary? Shouldn't it be none, feature, or hidden?

torch_frame/nn/models/excelformer.py
```diff
@@ -37,30 +39,55 @@ def feature_mixup(
         y (Tensor): The target.
         num_classes (int): Number of classes.
         beta (float): The concentration parameter of the Beta distribution.
+        mixup_type (str): The mixup methods.
+        mi_scores (Tensor): Mutual information scores for FEAT-MIX.
```
weihua916, line 43: can you add more detailed descriptions for these arguments?

torch_frame/nn/models/excelformer.py
```diff
@@ -37,30 +39,55 @@ def feature_mixup(
...
 
     Returns:
         x_mixedup (Tensor): The mixedup numerical feature.
         y_mixedup (Tensor): Transformed target of size
             :obj:`[batch_size, num_classes]`
     """
     assert num_classes > 0
+    assert mixup_type in ['ordinary', 'feature', 'hidden']
```
weihua916, line 51: where is 'none'?

torch_frame/nn/models/excelformer.py
```diff
@@ -152,79 +188,101 @@ def __init__(
...
+        if mixup_encoded:
+            assert tf.y is not None
+            x, y_mixedup = feature_mixup(
+                x,
+                tf.y,
+                num_classes=self.out_channels,
+                beta=self.beta,
+                mixup_type=self.mixup,
+                mi_scores=getattr(tf, 'mi_scores', None),
+            )
```
weihua916, line 226: Where did you set `mixup_encoded` to be `True`?

weihua916 (commented):
Thank you for the contribution!

Left some initial comments and questions. It'd be great to clean-up the logic and add more documentation of different mixup strategies, `none`, `feature`, `hidden`, `ordinary` and test all of them.

Personal opinion. Do we need to add all of them? If some works better than the others, I would prefer to add less of them to avoid confusion.

torch_frame/nn/utils/loss.py
```diff
@@ -0,0 +1,11 @@
...
+from torch import Tensor
+
+
+def cross_entropy_with_logits(
+    logits: Tensor,
+    targets: Tensor,
+) -> Tensor:
+    r"""Cross Entropy Loss for Soft Multi-classification Labels."""
+    probs = torch.softmax(logits, dim=1)
+    return -(targets * probs.log()).sum(1).mean()
```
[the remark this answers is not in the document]
jyansir: Thank you for the instant feedback!

Yes, I see, the `F.cross_entropy` and `CrossEntropyLoss` have supported the soft targets since PyTorch `1.10.0`, while I kept the habit of manually realizing it as in the old versions, I have removed the file `torch_frame/nn/utils/loss.py`, it is indeed unnecessary, thank you!

For four mixup types (`none`, `feature`, `hidden`, `ordinary`), the `ordinary` is the previous mixup version implemented in this project, thus I left that option in the codes. In the paper we propose `feature` & `hidden` and both are always better than `ordinary`. `none` represents no mixup, and sometimes there are few datasets preferring this option, it depends on inherent data characteristics. **As you suggested, I have removed the `ordinary` option**. Since other options are totally compatible with the advanced features of `torch.compile`, **I uniformly integrate them into `forward` function and remove `foward_mixup` which was for the `ordinary` option, if the user needs mixup during training, just simply calling `model(tf, mixup_encoded=True)`, and during inference without mixup, call `model(tf)`.**

The related documentation is detailed and fixed, and `torch.compile` testings are passed.

One more thing, I found my pull requests cannot pass the pytest under PyTorch `1.12.0`, `1.13.0` and `2.0.0` with a consistent single failed case here, while I configured the same environments as action files (I'm not sure the local ones are exactly the same as the used ones in GitHub) with all these versions no problem. The failed case here is `batch_size = 0`, `mixup = 'hidden'`, `task_type = 'binary_classification'`, it seems everything in real usage (`batch_size != 0`) or in my local machine (`batch_size == 0`) is all right.

examples/excelformer.py
```diff
@@ -20,17 +20,19 @@
 from torch_frame.data.loader import DataLoader
 from torch_frame.datasets.yandex import Yandex
 from torch_frame.nn import ExcelFormer
+from torch_frame.nn.utils.loss import cross_entropy_with_logits
```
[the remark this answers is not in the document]
jyansir: Yes, I have fixed it.

benchmark/data_frame_benchmark.py
```diff
@@ -27,6 +27,7 @@
     TabTransformer,
     Trompt,
 )
+from torch_frame.nn.utils.loss import cross_entropy_with_logits
```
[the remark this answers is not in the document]
jyansir: Yes, I have fixed it.

examples/excelformer.py
jyansir: As replied above, the `ordinary` is the previous mixup version implemented in this project, thus I left that option in the codes, while in our paper, the proposed `feature` & `hidden` are always better than `ordinary`. As suggested, I have removed `ordinary` and only keep the `none` & `feature` & `hidden`.

torch_frame/nn/models/excelformer.py
jyansir: Sure, I'm making this part detailed now!

torch_frame/nn/models/excelformer.py
jyansir: Thank you! I have fixed it with only  `none` & `feature` & `hidden` options.

torch_frame/nn/models/excelformer.py
jyansir: I have organized this in a more clear manner. Your mentioned `mixup_encoded = True` appeared in the `forward_mixup` function since I wanted to call `feature` & `hidden` mixup (mixup on encoded features) through `forward_mixup` API. Now I have re-organized the calling style, `forward_mixup` is removed (since we don't need `ordinary` mixup), and the user directly uses mixup for training by simply calling `model(tf, mixup_encoded=True)`.

torch_frame/nn/utils/loss.py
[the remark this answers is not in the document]
weihua916: Thank you for addressing the comments!

torch_frame/nn/models/excelformer.py
```diff
@@ -71,8 +102,8 @@ class ExcelFormer(Module):
 
     ExcelFormer first converts the categorical features with a target
     statistics encoder into numerical features. Then it sorts the
-    numerical features with mutual information sort. So the model
-    itself limits to numerical features.
+    numerical features with mutual information sort. For categorical
+    features, they are converted to numerical ones with CatBoostEncoder.
```
weihua916, line 106: nit: :class:`CatBoostEncoder`

torch_frame/nn/models/excelformer.py
```diff
@@ -37,30 +39,59 @@ def feature_mixup(
...
+    assert mixup_type in ['none', 'feature', 'hidden']
+
+    beta = torch.tensor(beta, dtype=torch.float32, device=x.device)
     beta_distribution = torch.distributions.beta.Beta(beta, beta)
     shuffle_rates = beta_distribution.sample(torch.Size((len(x), 1)))
-    feat_masks = torch.rand(x.shape, device=x.device) < shuffle_rates
     shuffled_idx = torch.randperm(len(x), device=x.device)
+    assert x.ndim == 3, 'FEAT-MIX or HIDDEN-MIX requires encoded features'
+    if mixup_type == 'feature':
+        assert mi_scores is not None
```
weihua916, line 64: Can we make this more informative error?

torch_frame/nn/models/excelformer.py
```diff
@@ -37,30 +39,59 @@ def feature_mixup(
...
+        mi_scores = mi_scores.to(x.device)
+        # Hard masks (feature dimension)
+        feat_masks = torch.rand(torch.Size((x.shape[0], x.shape[1])),
```
weihua916, line 67: nit: simply write as `torch.rand(x.shape, device=x.device)`

torch_frame/nn/models/excelformer.py
```diff
@@ -37,30 +39,59 @@ def feature_mixup(
...
+        feat_masks = torch.ones_like(x, dtype=torch.bool)
+        lmbd = torch.ones_like(shuffle_rates)
     x_mixedup = feat_masks * x + ~feat_masks * x[shuffled_idx]
 
     y_shuffled = y[shuffled_idx]
     if num_classes == 1:
         # Regression task or binary classification
-        shuffle_rates = shuffle_rates.view(-1, )
-        y_mixedup = shuffle_rates * y + (1 - shuffle_rates) * y_shuffled
+        lmbd = lmbd.squeeze(1)
```
weihua916, line 88: Can you clarify what `lmbd` stands for? :) Can there be a better name?

torch_frame/nn/models/excelformer.py
```diff
@@ -37,30 +39,59 @@ def feature_mixup(
...
+        # Hard masks (feature dimension)
+        feat_masks = torch.rand(torch.Size((x.shape[0], x.shape[1])),
+                                device=x.device) < shuffle_rates
+        l1_norm_mi = mi_scores / mi_scores.sum()
+        lmbd = torch.sum(
+            l1_norm_mi.unsqueeze(0) * feat_masks, dim=1, keepdim=True)
+        feat_masks = feat_masks.unsqueeze(2)
+    elif mixup_type == 'hidden':
+        # Hard masks (hidden dimension)
+        feat_masks = torch.rand(torch.Size((x.shape[0], x.shape[2])),
```
weihua916, line 75: same here. can be made simpler.
`torch.rand(x.shape, device=...)

weihua916 (approved):
LGTM!

torch_frame/nn/utils/loss.py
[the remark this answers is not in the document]
weihua916: > One more thing, I found my pull requests cannot pass the pytest under PyTorch 1.12.0, 1.13.0 and 2.0.0 with a consistent single failed case here, while I configured the same environments as action files (I'm not sure the local ones are exactly the same as the used ones in GitHub) with all these versions no problem. The failed case here is batch_size = 0, mixup = 'hidden', task_type = 'binary_classification', it seems everything in real usage (batch_size != 0) or in my local machine (batch_size == 0) is all right.

Apologies. That's a flaky test on our side. Fixed in https://github.com/pyg-team/pytorch-frame/pull/392

weihua916:
As a follow-up, it'd be great if you could run the updated https://github.com/pyg-team/pytorch-frame/blob/master/examples/excelformer.py and update the number https://github.com/pyg-team/pytorch-frame/blob/master/examples/excelformer.py#L5-L10

torch_frame/nn/models/excelformer.py
jyansir: Sure, I'm making it detailed now.

torch_frame/nn/models/excelformer.py
jyansir: As illustrated in the Fig. 3 of ExcelFormer [paper](https://arxiv.org/pdf/2301.02819.pdf), FEAT-MIX & HIDDEN-MIX apply mixup at feature dimension & hidden ones respectively, here the `x` is encoded features with 3 dimensions (batch size, feature amount, hidden size), uniformly writing as `torch.rand(x.shape, device=x.device) < shuffle_rates` in both `feature` and `hidden` options make their behaviors meaningless and different from the paper.

torch_frame/nn/models/excelformer.py
jyansir: Thank you! Yes, I'd like to change the mixup weight name to `lam` which is commonly used in mixup implementation, like repos of [mixup-cifar10-Line 119](https://github.com/facebookresearch/mixup-cifar10/blob/main/train.py#L119) and [Manifold-MixUp-Line 106](https://github.com/vikasverma1077/manifold_mixup/blob/master/supervised/models/resnet.py#L106). Necessary comment will be added for the variable.

torch_frame/nn/models/excelformer.py
jyansir: I think write as `b, f, d = x.shape` then `torch.rand(torch.Size((b, f)), device=x.device)` for `feature` option (and `torch.rand(torch.Size((b, d)), device=x.device)` for `hidden` option) can also seem readable and brief.

torch_frame/nn/models/excelformer.py
jyansir: As mentioned above, `feature` or `hidden` options only mixup at feature or hidden dimension, `torch.rand(x.shape, device=...)` makes inconsistent mixup weights in both dimensions that is not expected in the paper, for example, using `hidden` means each feature share the same mixup weights of hidden dimension. `torch.rand(x.shape, device=...)` indeed creates a mixture of `feature` & `hidden` while this mixup manner performs not good in our trail.

torch_frame/nn/models/excelformer.py
weihua916: Sounds good. Makes sense.

torch_frame/nn/models/excelformer.py
weihua916: Yes, `lam` is probably better. Thanks :)

jyansir:
Of course, I'd like to update the experimental results for ExcelFormer. Is there any required hardware or software setting? I'd like to run with PyTorch 2.2.0, Python 3.10, NVIDIA RTX 3090 in default. If specific hardware is required, mention it (4090 or A100 is available).

torch_frame/nn/models/excelformer.py
weihua916: makes sense.

weihua916:
Thanks! Any hardware is fine. It does not affect the results anyway.

torch_frame/nn/models/excelformer.py
```diff
@@ -37,30 +39,59 @@ def feature_mixup(
...
 
     Returns:
         x_mixedup (Tensor): The mixedup numerical feature.
         y_mixedup (Tensor): Transformed target of size
             :obj:`[batch_size, num_classes]`
     """
     assert num_classes > 0
+    assert mixup_type in ['none', 'feature', 'hidden']
+
+    beta = torch.tensor(beta, dtype=torch.float32, device=x.device)
```
yiweny, line 58: Maybe use the same dtype as the tensor?

torch_frame/nn/models/excelformer.py
```diff
@@ -37,30 +39,59 @@ def feature_mixup(
...
     shuffled_idx = torch.randperm(len(x), device=x.device)
+    assert x.ndim == 3, 'FEAT-MIX or HIDDEN-MIX requires encoded features'
+    if mixup_type == 'feature':
+        assert mi_scores is not None
+        mi_scores = mi_scores.to(x.device)
+        # Hard masks (feature dimension)
+        feat_masks = torch.rand(torch.Size((x.shape[0], x.shape[1])),
+                                device=x.device) < shuffle_rates
+        l1_norm_mi = mi_scores / mi_scores.sum()
+        lmbd = torch.sum(
```
yiweny, line 70: Better naming of the variables?

torch_frame/nn/models/excelformer.py
```diff
@@ -71,8 +102,8 @@ class ExcelFormer(Module):
 
     ExcelFormer first converts the categorical features with a target
     statistics encoder into numerical features. Then it sorts the
-    numerical features with mutual information sort. So the model
-    itself limits to numerical features.
+    numerical features with mutual information sort. For categorical
```
yiweny, line 105: I think it's better to keep the line that the model limits to numerical features and if you'd need categoricals you need to convert them first.

torch_frame/nn/models/excelformer.py
```diff
@@ -120,12 +157,16 @@ def __init__(
         diam_dropout: float = 0.0,
         aium_dropout: float = 0.0,
         residual_dropout: float = 0.0,
+        mixup: str = 'none',
```
yiweny, line 160:
```suggestion
        mixup: str | None = None,
```

torch_frame/nn/models/excelformer.py
```diff
@@ -120,12 +157,16 @@ def __init__(
...
+        beta: float = 0.5,
     ) -> None:
         super().__init__()
         if num_layers <= 0:
             raise ValueError(
                 f"num_layers must be a positive integer (got {num_layers})")
 
+        assert mixup in ['none', 'feature', 'hidden']
```
yiweny, line 168: Change here as well, as mixup can be `None`

torch_frame/nn/models/excelformer.py
```diff
@@ -37,30 +39,59 @@ def feature_mixup(
         y (Tensor): The target.
         num_classes (int): Number of classes.
         beta (float): The concentration parameter of the Beta distribution.
+            (default: :obj:`0.5`)
+        mixup_type (str): The mixup methods. `none` option means no mixup,
```
yiweny, line 43:
```suggestion
        mixup_type (str, optional): The mixup methods. `none` option means no mixup,
```
Please make mixup_type optional instead.

yiweny (commented):
Thanks! I left some initial comments

jyansir:
> Thanks! I left some initial comments

Thank you, yiwen. I will make the necessary adjustments based on the mentioned suggestions and get back to you.

torch_frame/nn/models/excelformer.py
jyansir: You' are right, I have fixed it.

torch_frame/nn/models/excelformer.py
jyansir: I have renamed `feat_masks`, `l1_norm_mi`, `lmbd` to `mixup_mask`, `norm_mi_scores`, `lam` respectively with comments for their meanings.

torch_frame/nn/models/excelformer.py
jyansir: Yes, I agree and have kept it.

torch_frame/nn/models/excelformer.py
jyansir: Sure, I have made `mixup_type` optional with supporting default `None` value mentioned above. The document has been adjusted accordingly.

jyansir:
I have made the necessary adjustments suggested by @yiweny and added experimental results for new ExcelFormer mentioned by @weihua916. Thank you for the beneficial suggestions!

examples/excelformer.py
```diff
@@ -23,14 +28,16 @@
 from torch_frame.transforms import CatToNumTransform, MutualInformationSort
 
 parser = argparse.ArgumentParser()
-parser.add_argument('--dataset', type=str, default='higgs_small')
+parser.add_argument('--dataset', type=str, default='california_housing')
+parser.add_argument('--mixup', type=str, default=None,
```
weihua916, line 32: `choices` should include None

examples/excelformer.py
```diff
@@ -23,14 +28,16 @@
...
+                    choices=['feature', 'hidden'])
 parser.add_argument('--channels', type=int, default=256)
 parser.add_argument('--batch_size', type=int, default=512)
 parser.add_argument('--num_heads', type=int, default=4)
 parser.add_argument('--num_layers', type=int, default=5)
 parser.add_argument('--lr', type=float, default=0.001)
+parser.add_argument('--gamma', type=float, default=0.95, choices=[0.95, 1.00])
```
weihua916, line 39: type=float
delete choices

examples/excelformer.py
```diff
@@ -76,6 +83,16 @@
...
 
+is_binary_class = is_classification and out_channels == 2
+
+if is_binary_class:
+    metric_computer = AUROC(task='binary')
+elif is_classification:
+    metric_computer = Accuracy(task='multiclass', num_classes=out_channels)
+else:
+    metric_computer = MeanSquaredError()
+metric_computer = metric_computer.to(device)
```
weihua916, line 94: Thank you for the update!

examples/excelformer.py
```diff
@@ -1,20 +1,25 @@
...
+The reproduced results are based on Z-score Normalization, and the
+reported ones are based on :class:`QuantileTransformer` preprocessing
+in the Sklearn Python package.
+
+california_housing: 0.4587 (0.4550) mixup: feature, num_layers: 3,
+gamma: 1.00, epochs: 300
+jannis : 72.51 (72.80) mixup: feature
+covtype: 97.17 (97.02) mixup: hidden
+helena: 38.20 (37.68) mixup: feature
+higgs_small: 80.75 (79.27) mixup: hidden
```
weihua916, line 14: Great results! Thank you.

examples/excelformer.py
```diff
@@ -1,20 +1,25 @@
...
-
-california_housing: 0.4587 (0.4733) num_layers=5, num_heads=4, num_layers=5,
-channels=32, lr: 0.001,
-jannis : 72.51 (72.38) num_heads=32, lr: 0.0001
-covtype: 97.17 (95.37)
-helena: 38.20 (36.80)
-higgs_small: 80.75 (65.17) lr: 0.0001
+The reproduced results are based on Z-score Normalization, and the
+reported ones are based on :class:`QuantileTransformer` preprocessing
+in the Sklearn Python package.
```
weihua916, line 7: What does this mean?

weihua916 (approved):
Thank you! Feel free to merge it!

examples/excelformer.py
jyansir: It seems the data preprocessing here is Z-score Normalization (i.e., `(x - x.mean) / x.std`, I find it in `ExcelFormerEncoder` [line 684 of `pytorch-frame/torch_frame/nn/encoder/stype_encoder.py`](https://github.com/pyg-team/pytorch-frame/blob/master/torch_frame/nn/encoder/stype_encoder.py#L684)), it also exists before encoding of other baselines. Actually, the preprocessing in ExcelFormer (and also in Yandex's FT-Transformer) is [`quantile`](https://github.com/yandex-research/rtdl-revisiting-models/blob/main/output/adult/ft_transformer/default/0/stats.json#L7) implemented with `QuantileTransformer` in the Sklearn Python package. I don't know if I miss some key observations in this project, while these two preprocessing methods are indeed different and affect the results.

Since all baselines here uniformly use the Z-score Normalization in their encoders, thus I think the experimental results of Z-score Normalization is OK, and I comment the different preprocessing in the example script.

If it is necessary to change different preprocessing methods, opening a PR in the future is more suitable.

weihua916: I see! Can you clarify that these normalization is over numerical feature?

jyansir: Of course! I have clarified the normalization is over numerical feature in the example script.

The modification has been committed, if there is no further problem, I'd like to resolve all conversations to merge them.

I'd like to contribute [T2G-Former](https://github.com/jyansir/t2g-former) here in the next month, if there is no near arrangement for adding the `quantile` preprocessing, I can finish this part at the same time.

weihua916: Thanks. Feel free to merge!!

jyansir: All conversations have been resolved 🤝, thank you for the beneficial suggestions of package development. I have no write access, you can merge it.

PR #397, opened 2024-05-10
Add `MovieLens 1M` dataset

xnuohz:


torch_frame/datasets/movielens_1m.py
```diff
@@ -0,0 +1,99 @@
...
+            names=['user_id', 'movie_id', 'rating', 'timestamp'],
+            sep='::',
+            engine='python',
+        )
+
+        df = pd.merge(pd.merge(ratings, users), movies) \
+               .sort_values(by='timestamp') \
+               .reset_index().drop('index', axis=1)
+        num_size = df.shape[0]
+        train_size, val_size = int(num_size * 0.8), int(num_size * 0.1)
```
yiweny, line 80: If the dataset doesn't come with a split, we don't need to create split here. PyTorch Frame has an easy interface to create splits.

xnuohz: Got it. Wonder if there is a simpler example that I can refer to divide the dataset in chronological order rather than randomly?

yiweny: We don't have an example we divide a dataset chronologically. I think the way to split the dataset can be user-defined and no need to be committed to the code base.

xnuohz: Got it. Already removed.

yiweny (approved):

PR #398, opened 2024-05-13
add pytorch tabular benchmark

yiweny:
This PR adds a scirpt to compare TabTransformer across pytorch_tabular and pytroch_frame. 
Results:

CPU:
-  Frame: 11.9 it/s 
- Tabular: 9.1 it/s

 
CUDA:
- Frame: 45.06 it/s
- Tabular: 41.76 it/s

 In a follow up PR add comparison for other models.


wsad1 (approved):

akihironitta (approved):

akihironitta (commented):
One thing to note is that since PyTorch Tabular uses PyTorch Lightning, I assume extra features are turned on by default, e.g., logging, checkpointing, sanity val steps, etc., making it unfair for pure comparison.

PR #402, opened 2024-05-22
modify double np.hstack to single np.ravel

February24-Lee:
```python
# before
np.hstack(np.hstack(x))

# after
np.stack(x)
```

When I tested the time with 880,591 x 1 series data,
the running time improved from almost 2.5 seconds to within 1.0 second.

I thought the usage of double np.hstack was for StateType.sequence_numerical. In this case, np.stack will be fine. But if there is something I missed, please let me know.


February24-Lee:
oh, there were something I missed. I will check and re-try to PR

akihironitta (commented):
Improving performance in the codebase is always welcome! Let us know if you need any help :)

PR #417, opened 2024-07-08
Towards supporting `MultiCategorical` encoder for target in torchframe

XinweiHe:
as titled.


torch_frame/data/tensor_frame.py
```diff
@@ -348,7 +348,6 @@ def _apply(self, fn: Callable[[TensorData], TensorData]) -> TensorFrame:
         out.feat_dict = {stype: fn(x) for stype, x in out.feat_dict.items()}
         if out.y is not None:
             y = fn(out.y)
-            assert isinstance(y, Tensor)
```
rusty1s, line 351: Let's relax the constraint to be of type `TensorData` rather than drop it. We still want to have methods such as `.cuda()` to be well defined and not error out.

akihironitta (approved):

PR #420, opened 2024-07-18
fix multicategorical stype inference and add test case

yiweny:


torch_frame/utils/infer_stype.py
```diff
@@ -138,18 +139,25 @@ def infer_series_stype(ser: Series) -> stype | None:
...
+                                    lambda row: MultiCategoricalTensorMapper.
+                                    split_by_sep(row, sep)).explode()))
+                    except Exception as e:
+                        logging.warn(
+                            "Mapping series into multicategorical stype "
+                            f"with separator {sep} raised an exception {e}")
+                        continue
+
+            if len(min_count_list) > 0 and max(
+                    min_count_list) > cat_min_count_thresh:
```
rusty1s, line 160: don't you need a unique here?

weihua916 (approved):

torch_frame/utils/infer_stype.py
```diff
@@ -138,18 +139,25 @@ def infer_series_stype(ser: Series) -> stype | None:
...
+                        min_count_list.append(
+                            _min_count(
+                                ser.apply(
+                                    lambda row: MultiCategoricalTensorMapper.
+                                    split_by_sep(row, sep)).explode()))
+                    except Exception as e:
+                        logging.warn(
+                            "Mapping series into multicategorical stype "
+                            f"with separator {sep} raised an exception {e}")
+                        continue
```
weihua916, line 157: nit: I would put `min_count_list` inside else:
then do `max(min_count_list)` in else statement. in the if-statement, I would just put the value of `_min_count(ser.explode())`.

Then judge if the value is larger than `cat_min_count_thresh`

yiweny: got it. Fixed

torch_frame/utils/infer_stype.py
yiweny: What do you mean by need a unique?

weihua916 (approved):

PR #421, opened 2024-07-30
coorectly infer boolean stypes

yiweny:


torch_frame/utils/infer_stype.py
```diff
@@ -109,6 +109,9 @@ def infer_series_stype(ser: Series) -> stype | None:
         # text_(embedded/tokenized)
 
         if ptypes.is_numeric_dtype(ser):
+
+            if ptypes.is_bool_dtype(ser):
+                return stype.categorical
```
akihironitta, line 114: It'd good to have a test case to cover this change.

akihironitta (approved):

torch_frame/utils/infer_stype.py
weihua916: +1

weihua916 (approved):
Thanks!

PR #423, opened 2024-07-31
Fixed warning and added safe globals

### LINKED ISSUE #422: Use `torch.load(weights_only=True)` in the codebase
It'd be nice to comply with this warning:
```
/home/aki/work/github.com/pyg-team/pytorch-frame/torch_frame/utils/io.py:98: FutureWarning: You are using `torch.load` with `weights_only=False` (the current default value), which uses the default pickle module implicitly. It is possible to construct malicious pickle data which will execute arbitrary code during unpickling (See https://github.com/pytorch/pytorch/blob/main/SECURITY.md#untrusted-models for more details). In a future release, the default value for `weights_only` will be flipped to `True`. This limits the functions that could be executed during unpickling. Arbitrary objects will no longer be allowed to be loaded via this mode unless they are explicitly allowlisted by the user via `torch.serialization.add_safe_globals`. We recommend you start setting `weights_only=True` for any use case where you don't have full control of the loaded file. Please open an issue on GitHub for any issues related to this experimental feature.
  tf_dict, col_stats = torch.load(path)
```

NeelKondapalli:
Fixes #422 by adding the `weights_only = True` argument to `torch.load` in the file `io.py`. This protects agains the arbitrary data warning. The types `stype` and `StatType` were added to the safe globals list.

By: Neel Kondapalli (neel2h06@gmail.com)


akihironitta (commented):
Thanks for quickly working on this issue! Looks like the CI is failing. Let me know if you need any help!

neelkkondapalli:
The fix works for a higher version of PyTorch, will investigate a fix for v2.2

NeelKondapalli:
The fix works for a higher PyTorch version, will investigate a fix for v2.2. Has to do with telling the code that the custom classes are safe for depickling.

NeelKondapalli:
@akihironitta It seems that the warning in #422  is not being reproduced for Torch v2.2.0 in the CI test logs. The warning occurs because pickle has a potential issue where unsafe code may be executed upon unpickling. Torch v2.4 is the only version where I was able to reproduce this warning. Torch v2.4 also contains the solution, being adding the custom classes to the safe list via the `torch.serialization.add_safe_globals` method. This method is not in v2.2, hence why CI failed. What is the best next step? Thanks!

https://github.com/pytorch/pytorch/pull/129239 is where the warning was introduced.

akihironitta (commented):
Sorry for the delay. Having a look today.

akihironitta (commented):
Looks great! Thanks again :)

I feel your 4ce8e5c21b21137d96b0ebbe614464704e7cfb49 is the right thing to do. Whenever possible, we should set `weights_only=True`, but we can disable it in other cases.

akihironitta (approved):

akihironitta:
@NeelKondapalli This is great! Thank you again for making your first contribution to PyTorch Frame! 🚀

neelkkondapalli:
Thank you! Awesome we could work it out.

PR #424, opened 2024-08-04
support xgboost early stopping

### LINKED ISSUE #419: XGBoost did not use the best one iteration to predict
Very good work!
But when I dug into the details of the implementation, I found a possible bug in your implementation.
From reading the official XGBoost documentation, it is clear that XGBoost.train returns the last iteration of the model, not the best one. More details can be checked [here](https://xgboost.readthedocs.io/en/latest/python/python_intro.html#early-stopping).
Therefore, I think the [code](https://github.com/pyg-team/pytorch-frame/blob/d81b8f7a9e0643fa553d7cb7a1343ef662fd6835/torch_frame/gbdt/tuned_xgboost.py#L175) below may be different from what was actually intended.

```python
boost = xgboost.train(
    self.params,
    dtrain,
    num_boost_round=num_boost_round,
    early_stopping_rounds=50,
    verbose_eval=False,
    evals=[(dvalid, "validation")],
    callbacks=[pruning_callback],
)
pred = boost.predict(dvalid)
```
The predicted code needs to be changed to:
```python
pred = boost.predict(dvalid, iteration_range=(0, boost.best_iteration + 1))
```
Similarly, other places used for prediction will likewise need to be modified.
I'm not sure if my understanding agrees with yours, and if it does, please fix this **as soon as possible**. The existence of this issue can lead to unfair comparisons.

yiweny:
Fixes #419

### DIFF
--- a/test/gbdt/test_gbdt.py
+++ b/test/gbdt/test_gbdt.py
@@ -54,12 +54,18 @@ def test_gbdt_with_save_load(gbdt_cls, stypes, task_type_and_metric):
         with pytest.raises(RuntimeError, match="is not yet fitted"):
             gbdt.save(path)
 
-        gbdt.tune(
-            tf_train=dataset.tensor_frame,
-            tf_val=dataset.tensor_frame,
-            num_trials=2,
-            num_boost_round=2,
-        )
+        if isinstance(gbdt_cls, XGBoost):
+            gbdt.tune(tf_train=dataset.tensor_frame,
+                      tf_val=dataset.tensor_frame, num_trials=2,
+                      num_boost_round=1000, early_stopping_rounds=2)
+            assert gbdt.model.best_iteration is not None
+        else:
+            gbdt.tune(
+                tf_train=dataset.tensor_frame,
+                tf_val=dataset.tensor_frame,
+                num_trials=2,
+                num_boost_round=2,
+            )
         gbdt.save(path)
 
         loaded_gbdt = gbdt_cls(
@@ -86,30 +92,3 @@ def test_gbdt_with_save_load(gbdt_cls, stypes, task_type_and_metric):
         assert (0 <= score <= 1)
     elif task_type == TaskType.MULTICLASS_CLASSIFICATION:
         assert (0 <= score <= 1)
-
-
-def test_xgboost_early_stopping_rounds():
-    dataset: Dataset = FakeDataset(
-        num_rows=30,
-        with_nan=True,
-        stypes=[stype.numerical, stype.categorical],
-        create_split=True,
-        task_type=TaskType.BINARY_CLASSIFICATION,
-        col_to_text_embedder_cfg=TextEmbedderConfig(
-            text_embedder=HashTextEmbedder(8)),
-    )
-    dataset.materialize()
-    gbdt = XGBoost(
-        task_type=TaskType.BINARY_CLASSIFICATION,
-        metric=Metric.ROCAUC,
-    )
-
-    gbdt.tune(
-        tf_train=dataset.tensor_frame,
-        tf_val=dataset.tensor_frame,
-        num_trials=2,
-        num_boost_round=100,
-        early_stopping_rounds=2,
-    )
-
-    assert gbdt.model.best_iteration is not None
--- a/test/gbdt/test_gbdt.py
+++ b/test/gbdt/test_gbdt.py
@@ -86,3 +86,30 @@ def test_gbdt_with_save_load(gbdt_cls, stypes, task_type_and_metric):
         assert (0 <= score <= 1)
     elif task_type == TaskType.MULTICLASS_CLASSIFICATION:
         assert (0 <= score <= 1)
+
+
+def test_xgboost_early_stopping_rounds():
+    dataset: Dataset = FakeDataset(
+        num_rows=30,
+        with_nan=True,
+        stypes=[stype.numerical, stype.categorical],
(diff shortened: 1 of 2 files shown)


test/gbdt/test_gbdt.py
```diff
@@ -86,3 +86,30 @@ def test_gbdt_with_save_load(gbdt_cls, stypes, task_type_and_metric):
...
+
+    gbdt.tune(
+        tf_train=dataset.tensor_frame,
+        tf_val=dataset.tensor_frame,
+        num_trials=2,
+        num_boost_round=100,
+        early_stopping_rounds=2,
+    )
+
+    assert gbdt.model.best_iteration is not None
```
weihua916, line 115: Can we include this test into the test above?

weihua916 (commented):
Thanks. The test seems failing. Also, let's mention about the issue.

torch_frame/gbdt/tuned_xgboost.py
```diff
@@ -174,10 +177,15 @@ def objective(
...
-                              callbacks=[pruning_callback])
-        pred = boost.predict(dvalid)
+                              early_stopping_rounds=early_stopping_rounds,
+                              verbose_eval=False, evals=[
+                                  (dvalid, 'validation')
+                              ], callbacks=[pruning_callback])
+        if boost.best_iteration:
+            iteration_range = (0, boost.best_iteration + 1)
+        else:
+            iteration_range = None
```
weihua916, line 187: will we ever enter this `else`?

torch_frame/gbdt/tuned_xgboost.py
```diff
@@ -225,7 +235,11 @@ def _predict(self, tf_test: TensorFrame) -> Tensor:
         dtest = xgboost.DMatrix(test_feat, label=test_y,
                                 feature_types=test_feat_type,
                                 enable_categorical=True)
-        pred = self.model.predict(dtest)
+        if self.model.best_iteration is not None:
+            iteration_range = self.model.best_iteration
+        else:
+            iteration_range = None
```
weihua916, line 241: ditto

torch_frame/gbdt/tuned_xgboost.py
zechengz: I think it's possible?
image

torch_frame/gbdt/tuned_xgboost.py
```diff
@@ -174,10 +177,15 @@ def objective(
...
                               num_boost_round=num_boost_round,
-                              early_stopping_rounds=50, verbose_eval=False,
-                              evals=[(dvalid, 'validation')],
-                              callbacks=[pruning_callback])
-        pred = boost.predict(dvalid)
+                              early_stopping_rounds=early_stopping_rounds,
+                              verbose_eval=False, evals=[
+                                  (dvalid, 'validation')
+                              ], callbacks=[pruning_callback])
+        if boost.best_iteration:
```
zechengz, line 184: Seems we missed this... https://xgboost.readthedocs.io/en/stable/python/python_api.html#xgboost.Booster.predict

test/gbdt/test_gbdt.py
yiweny: Done

torch_frame/gbdt/tuned_xgboost.py
yiweny: It's possible to hit this case, when there's no early stopping

torch_frame/gbdt/tuned_xgboost.py
yiweny: Yes, when there's no early stopping

torch_frame/gbdt/tuned_xgboost.py
```diff
@@ -93,6 +93,7 @@ def objective(
         dtrain: Any,  # xgboost.DMatrix
         dvalid: Any,  # xgboost.DMatrix
         num_boost_round: int,
+        early_stopping_rounds: int,
```
weihua916, line 96: nit: shall we set the default value to `50` for backward compatibility?

torch_frame/gbdt/tuned_xgboost.py
```diff
@@ -225,8 +240,19 @@ def _predict(self, tf_test: TensorFrame) -> Tensor:
...
+            iteration_range = None
+        pred = self.model.predict(dtest, iteration_range)
+
+        if (self.model.best_iteration
+                and self.task_type == TaskType.MULTICLASS_CLASSIFICATION):
+            assert pred.shape[1] == self._num_classes
+            pred = torch.argmax(torch.from_numpy(pred), dim=1)
+        else:
+            pred = torch.from_numpy(pred)
+        return pred.to(device)
```
weihua916, line 255: curious - why do we need to introduce these?

weihua916: Ok, now I understand. Do you mind adding a comment?

weihua916 (approved):
Thank you!

torch_frame/gbdt/tuned_xgboost.py
yiweny: Sure

torch_frame/gbdt/tuned_xgboost.py
yiweny: Yes. I set the `early_stopping_rounds` to default to 50

PR #427, opened 2024-08-07
Update Excelformer benchmark results on small binary and regression tasks

zechengz:
* Update Excelformer benchmark results on small binary and regression tasks
* Also update the README with arxiv icon


weihua916 (approved):
Thank you!

weihua916:
Would it be possible to do medium ones as a follow-up?

zechengz:
@weihua916 Yes I am running the medium and large experiments. Will update once they are finished.

PR #429, opened 2024-08-08
Adding support for openml datasets

qychen2001:
There are a large number of datasets available in openml, and this PR adds support for the openml dataset, enabling the integration of openml and torch_frame. The dataset used can be specified using dataset_id.


torch_frame/datasets/openml_dataset.py
```diff
@@ -0,0 +1,92 @@
...
+                assert df[target_col].nunique() == 2
+                self._task_type = torch_frame.TaskType.BINARY_CLASSIFICATION
+                self._num_classes = 2
+            else:
+                assert df[target_col].nunique(
+                ) == self.dataset_info["NumberOfClasses"]
+                self._task_type = (
+                    torch_frame.TaskType.MULTICLASS_CLASSIFICATION)
+                self._num_classes = int(self.dataset_info["NumberOfClasses"])
+            col_to_stype[target_col] = torch_frame.categorical
```
yiweny, line 65: Can you create a function to convert from openml `TaskType` to pytorch-frame `TaskType` and then call the function here?

torch_frame/datasets/openml_dataset.py
```diff
@@ -0,0 +1,92 @@
...
+            download_features_meta_data=True,
+        )
+        # Get dataset info from OpenML
+        self.dataset_info = self._openml_dataset.qualities
+        target_col = self._openml_dataset.default_target_attribute
+        X, y, self.categorical_indicator, _ = self._openml_dataset.get_data(
+            target=target_col)
+        df = pd.concat([X, y], axis=1)
+        self._task_type: torch_frame.TaskType = (
+            torch_frame.TaskType.BINARY_CLASSIFICATION)
```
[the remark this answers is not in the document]
qychen2001: The `_task_type` is distinguished from the `task_type` in `torch_frame.data.Dataset`, and is simply assigned an initial value here.
I have computed the real `task_type` for different datasets in the following code

```python
target_col_type = infer_series_stype(df[target_col])
if target_col_type == torch_frame.categorical:
    assert self.dataset_info["NumberOfClasses"] > 0
    if self.dataset_info["NumberOfClasses"] == 2:
        assert df[target_col].nunique() == 2
        self._task_type = torch_frame.TaskType.BINARY_CLASSIFICATION
        self._num_classes = 2
    else:
        assert df[target_col].nunique(
        ) == self.dataset_info["NumberOfClasses"]
        self._task_type = (
            torch_frame.TaskType.MULTICLASS_CLASSIFICATION)
        self._num_classes = int(self.dataset_info["NumberOfClasses"])
    col_to_stype[target_col] = torch_frame.categorical
else:
    assert self.dataset_info["NumberOfClasses"] == 0
    self._task_type = torch_frame.TaskType.REGRESSION
    self._num_classes = 0
    col_to_stype[target_col] = torch_frame.numerical
```

torch_frame/datasets/openml_dataset.py
```diff
@@ -0,0 +1,92 @@
+import os
+from typing import Optional
+
+import openml
```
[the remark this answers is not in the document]
qychen2001: I import openml in the class constructor and added the error handling

torch_frame/datasets/openml_dataset.py
qychen2001: I don't think that's a good idea. Because the `df` here would be used in the constructor of the parent class. Also `col_to_stype` needs to get done before the constructor. If I am going to use a function to do this part, then `df` needs to be created as a new member variable.

yiweny: I don't quite get it.

PR #432, opened 2024-08-10
Remove CUDA synchronizations by slicing input tensor with `int` instead of CUDA tensors in `nn.LinearEmbeddingEncoder`

akihironitta:
`start_idx` and `end_idx` used at `feat.values[:, start_idx:end_idx]` are on device, which leads to cuda synchronizations.

### DIFF
--- a/torch_frame/nn/encoder/stype_encoder.py
+++ b/torch_frame/nn/encoder/stype_encoder.py
@@ -716,10 +716,10 @@ class LinearEmbeddingEncoder(StypeEncoder):
     def init_modules(self) -> None:
         super().init_modules()
         num_cols = len(self.stats_list)
-        emb_dim_list = [stats[StatType.EMB_DIM] for stats in self.stats_list]
+        self.emb_dim_list = [stats[StatType.EMB_DIM] for stats in self.stats_list]
         self.weight_list = ParameterList([
             Parameter(torch.empty(emb_dim, self.out_channels))
-            for emb_dim in emb_dim_list
+            for emb_dim in self.emb_dim_list
         ])
         self.biases = Parameter(torch.empty(num_cols, self.out_channels))
         self.reset_parameters()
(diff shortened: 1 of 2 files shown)


torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -736,12 +738,12 @@ def encode_forward(
...
         x_lins: list[Tensor] = []
-        for start_idx, end_idx, weight in zip(feat.offset[:-1],
-                                              feat.offset[1:],
-                                              self.weight_list):
+        start_idx = 0
+        for idx, col_dim in enumerate(self.emb_dim_list):
+            end_idx = start_idx + col_dim
             # [batch_size, emb_dim] * [emb_dim, out_channels]
             # -> [batch_size, out_channels]
-            x_lin = torch.matmul(feat.values[:, start_idx:end_idx], weight)
```
yiweny, line 744: Curious, why is `@` better than `torch.matmul`?

yiweny (approved):
Good catch. Thank you!

torch_frame/nn/encoder/stype_encoder.py
zechengz: I think they are same?

torch_frame/nn/encoder/stype_encoder.py
```diff
@@ -736,12 +738,12 @@ def encode_forward(
...
+            x_lin = feat.values[:, start_idx:end_idx] @ self.weight_list[idx]
```
zechengz, line 746: Originally the synchronization will slow down the computation right?

zechengz (approved):

torch_frame/nn/encoder/stype_encoder.py
akihironitta: Yes!

PR #437, opened 2024-08-17
Support `pin_memory()` in `Multi{Embedding,Nested}Tensor` and `TensorFrame`

akihironitta:


yiweny (commented):
What is `pin_memory`? 🙂‍↔️

akihironitta (commented):
It's useful for moving data to GPU faster: https://pytorch.org/docs/stable/data.html#memory-pinning :)

torch_frame/data/multi_tensor.py
```diff
@@ -93,6 +93,12 @@ def cpu(self, *args, **kwargs):
     def cuda(self, *args, **kwargs):
         return self._apply(lambda x: x.cuda(*args, **kwargs))
 
+    def pin_memory(self, *args, **kwargs):
+        return self._apply(lambda x: x.pin_memory(*args, **kwargs))
+
+    def is_pinned(self) -> bool:
+        return self.values.is_pinned()
```
zechengz, line 100: Should this be
```suggestion
        return self.values.is_pinned() and self.offset.is_pinned()
```

zechengz (approved):
LGTM.
IMO usually we set pin_memory for the dataloader right?

akihironitta (commented):
> IMO usually we set pin_memory for the dataloader right?

Yes! The dataloader will now call this method on our multi tensors and tensor frame.

weihua916:
are we merging this? Why there are a lot of test failures?

akihironitta:
Will be merging this PR once the CI becomes green 🟢 Let me know if anyone has any review comments on this PR as I'm happy to address any concerns or elaborate on the change in this PR :)

test/data/test_multi_nested_tensor.py
```diff
@@ -95,8 +95,8 @@ def test_fillna_col():
                           torch.tensor([100], dtype=torch.float32)))
 
 
-@withCUDA
-def test_multi_nested_tensor_basics(device):
+@onlyCUDA
+def test_basics(device):
```
akihironitta, line 99: #476

PR #447, opened 2024-09-07
Add benchmark result from `ExcelFormer` on a large-scale multi-class classification task

akihironitta:


zechengz:
Thanks!

zechengz (approved):
Thanks!

PR #449, opened 2024-09-09
Improve Stats compute perf

Kh4L:
This PR improves performance of `StatType.compute` by calling `np.hstack` only when needed. 

Before change:
```
$ time python  gnn_node.py --dataset=rel-hm --task=user-churn --epochs 1 --max_steps_per_epoch 250
[...]
real    3m17.517s
user    3m19.252s
sys     0m6.757s
```

After change:

```
$ time python  gnn_node.py --dataset=rel-hm --task=user-churn --epochs 1 --max_steps_per_epoch 250
[...]
real    1m43.020s
user    1m45.716s
sys     0m6.022s
```

### DIFF
--- a/torch_frame/data/stats.py
+++ b/torch_frame/data/stats.py
@@ -85,7 +85,8 @@ class StatType(Enum):
         sep: str | None = None,
     ) -> Any:
         if self == StatType.MEAN:
-            flattened = np.hstack(np.hstack(ser.values))
+            val = np.hstack(ser.values) if ser.values.ndim > 1 else ser.values
+            flattened = np.hstack(val) if val.ndim > 1 else val
             finite_mask = np.isfinite(flattened)
             if not finite_mask.any():
                 # NOTE: We may just error out here if eveything is NaN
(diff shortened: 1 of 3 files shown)


weihua916:
Thanks.
@Kh4L could you pass the test? @akihironitta could you help review/update this PR?

torch_frame/data/stats.py
```diff
@@ -85,22 +85,25 @@ def compute(
         sep: str | None = None,
     ) -> Any:
         if self == StatType.MEAN:
-            flattened = np.hstack(np.hstack(ser.values))
+            val = np.hstack(ser.values) if ser.values.ndim > 1 else ser.values
+            flattened = np.hstack(val) if val.ndim > 1 else val
```
akihironitta, line 89: Looks like the chagne doesn't take account of the `sequence_numerical` case:
```
self = <StatType.MEAN: 'MEAN'>
ser = 0                                                 [nan]
1         [0.3309801272693331, nan, 0.3018088119575514]
2     ...42, nan, 0.9997390177887806, ...
19                            [nan, 0.36[1182](https://github.com/pyg-team/pytorch-frame/pull/449/checks#step:6:1183)4486676457]
Name: seq_num_1, dtype: object
sep = None

    def compute(
        self,
        ser: Series,
        sep: str | None = None,
    ) -> Any:
        if self == StatType.MEAN:
            val = np.hstack(ser.values) if ser.values.ndim > 1 else ser.values
            flattened = np.hstack(val) if val.ndim > 1 else val
>           finite_mask = np.isfinite(flattened)
E           TypeError: ufunc 'isfinite' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''

torch_frame/data/stats.py:90: TypeError
```
https://github.com/pyg-team/pytorch-frame/actions/runs/12376953024/job/34545299882

akihironitta (commented):
Hey @Kh4L, this PR looks great so far! 🔥 Would you still be interested in finishing this PR?

JiwaniZakir (commented):
The two-line flattening logic is now duplicated across three branches (`MEAN`, `STD`, `QUANTILES`), which makes future changes error-prone. Since this pattern is identical in all three cases, it should be extracted into a small helper (e.g., `_flatten_series(ser)`) and called once per branch.

The `ndim > 1` guard on `val` after the first `hstack` deserves more scrutiny. If `ser.values` is a 2D array (shape `N×M`), `np.hstack` on a single 2D array returns it unchanged, so `val` would still be 2D and the second `np.hstack(val)` would also return a 2D array — `flattened` would never actually be 1D in that case. The original double-hstack had the same flaw, but it's worth confirming whether a `ravel()` or `flatten()` call would be more semantically correct here than chained `hstack` checks.

It would also be worth adding a unit test specifically for the scalar (1D `ser.values`) case that this change is meant to handle, to prevent regression if the branching logic is ever touched again.

akihironitta (approved):

akihironitta (approved):
@Kh4L Thanks, Serge!

PR #451, opened 2024-09-11
Add more whitelisted variables for torch safe loading

Kh4L:
Needed types as we also serialize np ndarrays

### DIFF
--- a/torch_frame/__init__.py
+++ b/torch_frame/__init__.py
@@ -27,10 +27,16 @@ import torch_frame.gbdt  # noqa
 
 if WITH_PT24:
     import torch
+    import numpy as np
+    import _codecs
 
     torch.serialization.add_safe_globals([
         stype,
         torch_frame.data.stats.StatType,
+        np.core.multiarray.scalar,
+        np.dtype,
+        type(np.dtype(np.int32)),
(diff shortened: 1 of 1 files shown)


JiwaniZakir (commented):
The `type(np.dtype(np.int32))` entry on line ~35 evaluates to `numpy.dtype` at import time, making it identical to the `np.dtype` entry directly above it — one of these is redundant and should be removed.

`np.core.multiarray.scalar` references the deprecated `np.core` namespace, which was removed in NumPy 2.0 and replaced by `numpy._core`. This will raise an `AttributeError` for users on NumPy ≥ 2.0; it would be worth either guarding this with a version check or using `np._core.multiarray.scalar` (with a fallback for older NumPy).

Adding `_codecs.encode` to the safe globals deserves a comment explaining why it's necessary. `_codecs.encode` is a low-level built-in that can be used to construct arbitrary byte strings, and whitelisting it in the deserialization context weakens the security guarantees that `add_safe_globals` is meant to provide. At minimum, a comment referencing the specific saved artifact that requires it (e.g., a particular dtype or object serialized by an older version) would help reviewers and future maintainers assess the risk.

PR #454, opened 2024-09-18
Removing the deprecated `categorical_feature` parameter from `lightgbm.train(...)` function calls.

drivanov:
Using `categorical_feature` parameter in `lightgbm.Dataset()` instead of `lightgbm.train(...)` eliminates the following warnings:
```
test/gbdt/test_gbdt.py: 60 warnings
  /usr/local/lib/python3.10/dist-packages/lightgbm/engine.py:187: LGBMDeprecationWarning: Argument 'categorical_feature' 
to train() is deprecated and will be removed in a future release. Set 'categorical_feature' when calling lightgbm.Dataset() 
instead. See https://github.com/microsoft/LightGBM/issues/6435.
    _emit_dataset_kwarg_warning("train", "categorical_feature")
```


yiweny (approved):
Thanks a lot for your contribution!

PR #455, opened 2024-09-20
Fix offset in LinearEmbeddingEncoder

toenshoff:
This PR fixes an issue in the `LinearEmbeddingEncoder` class. The forward function does not increment the `start_idx` variable, so only the first one is used correctly if multiple embedding columns are present.

This probably affects the results in the [RelBench](https://arxiv.org/pdf/2407.20060) paper, so I suggest double-checking those.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- Fixed offset in `LinearEmbeddingEncoder` [#455](https://github.com/pyg-team/pytorch-frame/pull/455)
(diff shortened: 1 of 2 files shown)


akihironitta (approved):
Great catch!

I removed the changelog entry as this bug introduced in #432 hasn't appeared in any releases yet.

akihironitta:
> This probably affects the results in the [RelBench](https://arxiv.org/pdf/2407.20060) paper, so I suggest double-checking those.

Thank you for sharing your suggestion here. The paper uses an older version of PyTorch Frame, so there's no need to double-check the results in the paper.

akihironitta:
cc @yiweny for visibility in case you're running some benchmarks with master.

PR #474, opened 2024-12-27
Add an example of training a tabular model on multiple GPUs

akihironitta:
Adds an example for training a model on multiple GPUs.

#### Usage

```console
$ python examples/trompt_multi_gpu.py
```

#### Changes
To enable it, this PR modifies `Trompt` accordingly:
```diff
- out = Trompt(...).forward_stacked(tf)  # forward_stacked is removed
+ out = Trompt(...)(tf)
  assert out.size() == (batch_size, num_layers, num_channels)
```

#### Some highlights in the script
* Unlike `examples/trompt.py`, it computes training accuracy batch-wise instead of computing it at the end of each epoch to save time (although model parameters change over steps within each epoch).
* It reduces losses and metrics with `all_reduce` and `torchmetrics`'s API across all ranks at the end of each epoch.
* It avoids unnecessary device synchronisations within each epoch, e.g., by not calling `float(cuda_tensor)`, and by setting `repeat_interleave(..., output_size=...)`.
* ... (I'm happy to elaborate if anything in the script is unclear.)

#### Benchmark results from `--dataset jannis` on `g6.12xlarge` with four L4 GPUs

| | four GPUs | single GPU |
|-|-|-|
| time per training step | **0.725 seconds** | 0.741 seconds |
| time per training epoch | **29 seconds** | 117 seconds |
| test accuracy | 80.23 % | 80.29 % |

### DIFF
--- a/torch_frame/nn/models/trompt.py
+++ b/torch_frame/nn/models/trompt.py
@@ -122,7 +122,7 @@ class Trompt(Module):
             trompt_conv.reset_parameters()
         self.trompt_decoder.reset_parameters()
 
-    def forward_stacked(self, tf: TensorFrame) -> Tensor:
+    def forward(self, tf: TensorFrame) -> Tensor:
         r"""Transforming :class:`TensorFrame` object into a series of output
         predictions at each layer. Used during training to compute layer-wise
         loss.
@@ -152,6 +152,3 @@ class Trompt(Module):
         # [batch_size, num_layers, out_channels]
         stacked_out = torch.cat(outs, dim=1)
         return stacked_out
-
-    def forward(self, tf: TensorFrame) -> Tensor:
-        return self.forward_stacked(tf).mean(dim=1)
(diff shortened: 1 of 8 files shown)


torch_frame/nn/models/trompt.py
```diff
@@ -122,7 +122,7 @@ def reset_parameters(self) -> None:
             trompt_conv.reset_parameters()
         self.trompt_decoder.reset_parameters()
 
-    def forward_stacked(self, tf: TensorFrame) -> Tensor:
```
weihua916, line 125: Let's make sure this change does not break the example code.

akihironitta: Confirmed the change doesn't break these scripts across all supported task types:
```
examples/trompt.py
benchmark/data_frame_benchmark.py
benchmark/data_frame_text_benchmark.py
```

PR #496, opened 2025-02-09
Support Python 3.12 and Python 3.13

akihironitta:
* Adds support for Python 3.12 and Python 3.13.
* Disables tests that use `catboost` as they don't support Python 3.13 yet as reported in https://github.com/catboost/catboost/issues/2748.
* Bumps the PyTorch version required for `torch.compile` tests as PyTorch 2.6 is the first version that supports `torch.compile` with Python 3.13.

### DIFF
--- a/test/gbdt/test_gbdt.py
+++ b/test/gbdt/test_gbdt.py
@@ -1,4 +1,5 @@
 import os.path as osp
+import sys
 import tempfile
 
 import pytest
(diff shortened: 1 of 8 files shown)


test/gbdt/test_gbdt.py
```diff
@@ -13,7 +14,13 @@
...
 
 @pytest.mark.parametrize('gbdt_cls', [
-    CatBoost,
+    pytest.param(
+        CatBoost,
+        marks=pytest.mark.skipif(
+            sys.version_info >= (3, 13),
+            reason="Not supported on Python 3.13",
+        ),
+    ),
```
akihironitta, line 23:
```suggestion
    # TODO: Run CatBoost test on Python 3.13 once supported
    # https://github.com/catboost/catboost/issues/2748
    pytest.param(
        CatBoost,
        marks=pytest.mark.skipif(
            sys.version_info >= (3, 13),
            reason="Not supported on Python 3.13",
        ),
    ),
```

PR #523, opened 2025-04-14
Support CatBoost in Python 3.13

akihironitta:
https://github.com/catboost/catboost/releases/tag/v1.2.8 started supporting Python 3.13 via https://github.com/catboost/catboost/issues/2748.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -43,12 +43,3 @@ runs:
     - name: List installed packages
       run: pip list
       shell: bash
-
-    # TODO: Include catboost in Python 3.13 CI when catboost supports it:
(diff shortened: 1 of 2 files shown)


akihironitta:
Note to myself:
```
A module that was compiled using NumPy 1.x cannot be run in
NumPy 2.0.2 as it may crash. To support both 1.x and 2.x
versions of NumPy, modules must be compiled with NumPy 2.0.
Some module may need to rebuild instead e.g. with 'pybind11>=2.12'.

If you are a user of the module, the easiest solution will be to
downgrade to 'numpy<2' or try to upgrade the affected module.
We expect that some modules will need time to support NumPy 2.
```

PR #602, opened 2026-03-28
Add support for `pandas>=3.0.0`

akihironitta:
* Remove the `pandas<3.0.0` pin in `pyproject.toml`
* Fix `MultiCategoricalTensorMapper` in `torch_frame/data/mapper.py` to support pandas 3.0, which uses `StringDtype` instead of `object` for string columns and `pd.NA` instead of `np.nan` for missing values (pandas-dev/pandas#54792)

### DIFF
--- a/torch_frame/typing.py
+++ b/torch_frame/typing.py
@@ -14,6 +14,8 @@ from torch_frame.data.multi_nested_tensor import MultiNestedTensor
 WITH_PT20 = int(torch.__version__.split('.')[0]) >= 2
 WITH_PT24 = WITH_PT20 and int(torch.__version__.split('.')[1]) >= 4
 
+WITH_PD3 = int(pd.__version__.split('.')[0]) >= 3
+
 
 class Metric(Enum):
     r"""The metric.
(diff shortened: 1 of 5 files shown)


torch_frame/typing.py
```diff
@@ -14,6 +14,8 @@
 WITH_PT20 = int(torch.__version__.split('.')[0]) >= 2
 WITH_PT24 = WITH_PT20 and int(torch.__version__.split('.')[1]) >= 4
 
+WITH_PD3 = int(pd.__version__.split('.')[0]) >= 3
+
```
akihironitta, line 17:
```suggestion
WITH_PD3 = int(pd.__version__.split('.')[0]) >= 3
```

torch_frame/typing.py
```diff
@@ -13,7 +13,7 @@
 
 WITH_PT20 = int(torch.__version__.split('.')[0]) >= 2
 WITH_PT24 = WITH_PT20 and int(torch.__version__.split('.')[1]) >= 4
-
+WITH_PD3 = int(pd.__version__.split('.')[0]) >= 3
```
akihironitta, line 16:
```suggestion
WITH_PD3 = int(pd.__version__.split('.')[0]) >= 3

```

